PR #1, opened 2021-01-09
Transfer code from CourtListener

mattdahl:
This PR copies the relevant code from CourtListener's citations module with minimal substantive changes. Cf. https://github.com/freelawproject/courtlistener/issues/1521.

Key points:
- References to the Django DB have been replaced with calls to the `courts_db` package (cf. https://github.com/freelawproject/courts-db/pull/12)
- The `as_html()` methods have been removed from the `Citation` objects as they serve no purpose outside of CL
- To make `find_citations.py` easier to understand, I moved some of the helper functions to a new file (`helpers.py`), but no logic was changed
- Typing has been added to all functions
- Rather than importing `get_visible_text()` from `juriscraper` (https://github.com/freelawproject/juriscraper/blob/master/juriscraper/lib/html_utils.py#L163), I just copied it into `utils.py`. I thought this made sense because it's the only function that we need from `juriscraper`, and so it seemed silly to add that as a dependency just for that.
- The relevant tests were also copied from CL, and they all pass
- What is the purpose of this function: https://github.com/freelawproject/courtlistener/blob/master/cl/citations/reporter_tokenizer.py#L19? It is never actually invoked in CL, but I copied it over here anyway. Should we be calling that somewhere?

More changes are probably necessary to prepare this for publishing, but I think I've transferred the bulk of the code from CL in a working fashion.


mattdahl:
This is still a WIP, as apparently there are several typing errors I need to fix. (Has Python become Java???) I will push some additional, squashed commits when I fix that.

mlissner:
All your points sound good. Your one q:

> What is the purpose of this function: https://github.com/freelawproject/courtlistener/blob/master/cl/citations/reporter_tokenizer.py#L19? It is never actually invoked in CL, but I copied it over here anyway. Should we be calling that somewhere?

I don't know, which means it can be deleted. I checked the git blame for it, and it's all me, and it looks like my style, but....I guess whatever it was supposed to be never happened. We can only hope my git skills have improved.

Sorry about the type checking. We started doing it during the move to py3, and I've found it to be *extremely* helpful. The checker catches bugs regularly (especially around bytes vs strs), and it helps me to actually understand other people's code, since I'm often reviewing something I've never seen before. I've also noticed that making myself and others do the typing forces people to do simpler functions. If the typing is so horrible that you can't stand it, well, you might simplify your function. I once swore off java. I still think I'd never touch it. But...the typing I enjoy so far.

mattdahl:
I actually first learned to code in Java, worshiping at the altar of static typing, polymorphism, etc. etc., and I thought I had left all that behind... I guess I can see why type hinting is useful for Python too, but it's really annoying lol.

Do we even enforce the type checking in CL yet? When I run `mypy` there I get tons of errors. I have fixed them all in this repo now though.

Currently, the problem is that the `pylint` reporting task is not passing. Can you advise as to what am I supposed to do to fix that? When I run it locally, it generates a file with some suggested refactors, but surely I don't have to implement all of its suggestions?

mlissner:
> it's really annoying lol.

Well, we really don't want to annoy our volunteers unless it gives us a solid improvement in code quality. I think typing does that, but if there are better ways to handle this, I'm happy to think about it. You're the first to use our new project template since we added typing. It doesn't surprise me that there are some issues. There must be a better way to do this. Maybe @drewsilcock has thoughts.

> Do we even enforce the type checking in CL yet? When I run mypy there I get tons of errors. I have fixed them all in this repo now though.

No, not yet. I think we should turn it on though in some form b/c the current state of things is that we're adding typing to various places, but mypy isn't doing anything. It'd be better to check them as we add them, I suspect.

> Currently, the problem is that the pylint reporting task is not passing. Can you advise as to what am I supposed to do to fix that? When I run it locally, it generates a file with some suggested refactors, but surely I don't have to implement all of its suggestions?

I don't know what's normal here. I poked through and fixed a bunch of these, so things are better, but there are some that look really difficult to fix without introducing bugs. Some of the ones I fixed were boring, others made the code better. Maybe the easiest way to deal with the remaining ones is to ignore them with `#pylint` comments in the code. That way we can keep the linting and express our aspiration to someday fix the old code (and it doesn't have to be fixed now).

What do you think?

mlissner:
Oh, one other thought: It'd be nice to have some check that says, "Did pylint get worse? Did mypy get worse?" And if so reject it, but if not allow it. I don't know how to set that up though....

drewsilcock:
There are bots that tell you whether code coverage got better or worse, maybe there's one for telling you whether pylint / mypy got better or worse? The other option is just to use mypy to raise warnings but not fail builds, and then a human reviewer can choose whether typings should be added or not. You could also raise the typing failures as annotations on the review using GitHub Action magic (I'm not sure how easy this would be).

The pylint report is annoying - I didn't know that when pylint writes a report it returns non-zero exit code when warnings are issued, with the exit code representing a bitmask of what types of errors / warnings were issued (???). I'll put a fix in for this.

mlissner:
I found that pylint has a parameter you can use to set the threshold and I just added it. Seems like an ok fix. I didn't find anything that'd enforce improvement.

Now the tests are failing...

jcushman:
In case it's helpful, I enjoyed this article [Why Pylint is both useful and unusable, and how you can actually use it](https://pythonspeed.com/articles/pylint/). The author recommends turning on pylint checks in small batches and then turning off ones that are mostly noise, to find the ones that are going to have high enough signal-to-noise ratio in your project to be useful. I've been tempted but haven't actually turned on pylint for any projects, so can't say how this pans out.

mlissner:
> This just generates the HTML report so it seems fine to me to just add --exit-zero. There's a separate job that does pylint -E to fail on errors but not warnings.

OK, I was wondering what was going on there since I didn't recognize the short flag offhand. I think I'll crank that one up so it dies if the threshold is below 9 or so, and so it includes warnings too. I think @jcushman's blog points to the right way. We're actually in pretty good shape here. I do shiver at ever getting this turned on in CL though. I guess I've got my work cut out for me (I wouldn't want anybody else to have to do all of CL).

mlissner:
OK, all tests are passing. To get this working, I tweaked the lint CI file and I removed support for python 3.6. Hopefully that's OK. It's scheduled for deprecation at the end of this year, so this is perhaps a bit aggressive, but 3.6 doesn't have support for `re.Match`, which we used for typing. We could hack around it, but it seemed easier to just remove support until somebody needs it.

@mattdahl is there anything else outstanding before I do a full review?

jcushman:
3.7 is fine with me. It's the version we have in production, being the
python version on Debian stable (Buster) which was released 18 months ago.

jcushman:
I have this installed locally alongside our (CAP's) cite extractor, and it's great to see it working! 🎉

I'm going to drop some feedback here as I mess around -- please don't hold pulling this on my account, though. I'm happy to refile stuff as bugs if it doesn't end up making sense to fold into this initial PR. Just want to share now in case it's handy to make changes along the way.

First thing I've noticed is the default behavior of `get_citations()` with `disambiguate=True` is surprising. Right now by default it returns only unambiguous citations, so for example `get_citations("23 Wall. 307") == []`. If you include `disambiguate=False` then all citations are returned but no disambiguation is done. At a minimum I'd switch the default.

What might be more useful is if the default behavior is for each cite to come back with a list of all the reporters in reporters_db that they potentially match (including reporter metadata like start and end date), so callers can do their own disambiguation with more context, and the `disambiguate=True` flag causes `get_citations` to attempt to narrow that list down to one but not filter anything out.

I think it'll turn out there's an elegant way to implement this internally by having `tokenize()` return `class ReporterStr(str)` subclasses instead of `str`s, that can also have a list of candidate reporters attached, and then use those when scanning the tokens instead of re-checking the master list of reporter strings. This also lets CAP do OCR-friendly tokenization just by plugging the tokenize() function. I can send this part as a separate PR later if helpful.

jcushman:
(Again just recording notes for later use, don't mind me.)

I did a bit of speed profiling, which is easier with this code separated out, running `py-spy record -o profile.svg -- python profile.py` on a script that just extracts cites from a large case ten times in a loop.

Looks like 61% of runtime goes to the line `if citation_token in list(EDITIONS.keys()) + list(VARIATIONS_ONLY.keys()):`, which can be eliminated by caching `REPORTER_STRINGS = set(list(EDITIONS.keys()) + list(VARIATIONS_ONLY.keys()))` earlier in the file.

20% goes to `elif strip_punct(citation_token.lower()) == "supra":`, which can be substantially reduced by replacing `strip_punct` with a less fancy regex like `re.sub(r'[^a-z]', '', citation_token.lower())`.

10% (or 50% once those are fixed) goes to `strings = REPORTER_RE.split(text)` in `tokenize()`, because of the giant `"|".join()` regex. I think we can eliminate the `tokenize()` runtime either with an optional hyperscan dependency or by doing more post-processing on a simpler regex, but that one will take a little more thought.

So, good news -- there should be an easy 5x speedup here from tweaking a couple of lines, and then another 2x or so by messing with tokenize after that.

mlissner:
I literally slapped my forehead reading those performance notes, @jcushman. Those feel like a nice opportunity for a lean PR once this one is merged. Hyperscan kind of freaks me out, but making it optional seems fine. We're not using Intel processors ourselves...I wonder if it'd still be faster (we are x86).

I'm down generally with the API changes you're describing in the disambiguation code. The current code is very much "Mike working in a vacuum," so I expect some low-hanging fruit in ergonomics.

Thanks for the comments.

mattdahl:
@mlissner I don't think anything else remains on my end. Substantively, I think this should be all the code that merit copying from CL.

Thanks for fixing all the linting issues. And great work on those performance calculations, @jcushman!

Once this is merged I also see some areas where I'd like to clean up the code a bit (if only for my own sanity)...

mlissner:
LGTM. Some next steps from the thread:

1. merge into CL, yes, duh
2. Fix biggest performance issues
3. Clean up code to keep Matt sane
4. Get a deploy going (see #3)

mlissner:
Thank you everybody!

PR #5, opened 2021-01-14
Factor out text cleaning

jcushman:
There's currently a few different text cleaning things happening that I know of:

* html is automatically stripped by default
* tokenizer collapses repeated spaces with `text = re.sub(" +", " ", text)`
* tokenizer removes repeated underscores with 
  ```
      # get rid of all the annoying underscores in text from pdfs
      text = re.sub(r"__+", "", text)
  ```

This stuff seems likely to be caller-specific, e.g. CAP will want different things from CL. I'm offering this PR for discussion as one way to refactor. Here's what it includes:

* Html now defaults to False. I like the idea of including this in the library because it's finicky to get right, but I'd want to opt in if I knew I had HTML.
* Collapsing spaces is a good default, and makes sense as a separate pass before the text goes to tokenize(). Now runs as a separate function behind a flag defaulting to True.
* Remove underscore replacing. This strikes me as specific to a particular text processing pipeline (and could conceivably even cause weirdness when double underscores are used as placeholder page numbers), so better for the caller to handle instead of living in the library.

This PR also includes a little refactor of the test harness to get the tests working with the new html default, so test pairs can have individual kwargs like `html=True`.

### DIFF
--- a/microscope/cleaners.py
+++ b/microscope/cleaners.py
@@ -1,0 +1,42 @@
+import re
+from typing import Callable, Dict
+
+import lxml.html
+
+
+def html(html_content: str) -> str:
+    """Given HTML markup, return only text that is visible
+    Adopted from freelawproject/juriscraper/lib/html_utils.py#L163
+
+    :param html_content: The HTML string
+    :return: Text that is visible
+    """
+    html_tree = lxml.html.fromstring(html_content)
+    text = html_tree.xpath(
+        """//text()[normalize-space() and not(
+            parent::style |
+            parent::link |
+            parent::head |
+            parent::script)]"""
+    )
+    return " ".join(text)
+
+
+def whitespace(text: str) -> str:
+    """Collapse multiple spaces into one, and strip whitespace."""
+    text = re.sub(" +", " ", text)
+
+    return text.strip()
+
+
+def underscores(text: str) -> str:
+    """Remove strings of two or more underscores that are common
+    in text extracted from PDFs."""
+    return re.sub(r"__+", "", text)
+
+
+cleaners_lookup: Dict[str, Callable[[str], str]] = {
+    "html": html,
+    "whitespace": whitespace,
+    "underscores": underscores,
+}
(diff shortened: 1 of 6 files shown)


mlissner:
It sort of feels like `get_citations` is doing too much. Like, maybe we just need another function or a class called `clean_text` or `TextCleaner` that lives outside of that function. It could then get the configuration of whatever cleaners we want turned on. So you could do, say:

```py3
text = clean_text(remove_html=True, elide_underscores=True, basic_cleanups=True)
get_citations(text)
```

I guess if we wanted to make it a function, we could allow `kwargs` too, so that we can add more cleaners for folks to enable down the road. Doing it as a class might be a closer fit with lxml's approach, but I almost always grab a function out of my toolbox for whatever reason.

I like that it's a single call to get_citations right now, but as we start adding questions about html, whether and how to clean text, etc, it gets messy. I'm also thinking that if we pull out the underscore stuff, I'm going to have the above pattern in CourtListener code anyway:

```py3
text = elide_underscores(text)
get_citations(html=True, clean_text=True)
```

I defer to other re defaults. My priors are too strong to trust myself with on that.

jcushman:
I like your `clean_text` suggestion. Or an API like this would also work:

```
get_citations(text, clean=['html', 'underscores', my_func, 'whitespace'])
```

Implemented as:

```
def get_citations(text, ..., clean=['whitespace']):
    for clean_step in clean:
        if callable(clean_step):
            text = clean_step(text)
        else:
            text = getattr(microscope.cleaners, clean_step)
```

So vanilla usage on text is `get_citations(text)`. CL's usage is `get_citations(text, clean=['html', 'underscores', 'whitespace'])`. If a caller needs to clean something after stripping html but before compacting whitespace, they can drop in a callable. And any built-in cleaning steps just go in as functions in cleaners.py.

mlissner:
I like it. Simple and flexible enough.

jcushman:
OK, should be good to go. Had to add a couple of `# type: ignore` cause mypy can't quite track that a string always comes out, though I think the type hints ought to let it.

mlissner:
Very nice, thanks.

PR #6, opened 2021-01-15
(cleanup) drop header lines

jcushman:
I believe we don't need file encoding lines in python 3 since the default is utf8, and don't need shebangs for files that don't make sense to mark executable. Just a quick cleanup PR -- no big deal to close if you want to keep these around for whatever reason.


mlissner:
I swear I yanked these from CL not too long ago. Py2....it pulls you back in.

PR #10, opened 2021-01-19
Tokenizer emits token objects instead of strings

jcushman:
I apologize that this is a large PR, both as far as being hard to review and being a big change to propose. Happy to back things out if they don't seem right, or to otherwise help with review.

## Intro

The core idea here is it would be nice if I could plug in a different tokenizer that has a different idea of what a reporter string is. In particular, because CAP comes from OCR, we have a certain amount of vagueness about distinguishing periods, commas, dashes, single-quotes, and specks of dust. So we'd like to recognize "1 U.'S, 1" as a citation, but still take advantage of the disambiguation and other stuff eyecite does.

How do we do that? Suppose instead of the tokenizer emitting `["1", "U.S.", "1"]`, it emits `[Token("1"), ReporterToken("U.S.", editions=[us_edition]), Token("1")]` -- that is, it's the tokenizer's job not just to say "this chunk is a reporter," which it does already, but also "these are the editions it might be referring to." The post-processing code then just checks that attribute instead of doing a new lookup based on the matched string, and you can drop in different tokenizers and everything works.

This complicates things a bit internally, but less than you would think. Downside is we add a class hierarchy of tokens, and have to have the tokenizer wrap things in `Token` and the post-processing code call `str()` when it needs plain strings. But it simplifies things too, because we can just calculate things once per token instead of looking it up over and over. So my hope is in addition to solving CAP's usecase (and other usecases like allowing the caller to plug in hyperscan if they want), this also pays for itself in simplicity. At least lines of code didn't increase much. :)

## Details

Specific changes in the diff:

* `get_citations()` now takes a `tokenizer` function that defaults to the built-in `tokenize`. See the test for an example.
* Add two models: `Reporter` that maps to the top-level reporters in reporters-db, like the set of "S.W" editions, and `Edition` that maps to "S.W.", "S.W.2d" etc. Each Edition points to its Reporter.
  * Are those the correct names for those concepts? Reporter means a million things here.
* In reporter_tokenizer.py, build an `EDITIONS_LOOKUP` from official reporter string to Edition object, and `VARIATIONS_LOOKUP` from variation string to Edition object. We have to keep these separate so we can preferentially resolve ambiguous reporters to an exact match instead of a variation.
* In reporter_tokenizer.py, define classes for each kind of token we check for later in the code -- Token, ReporterToken, SectionToken, SupraToken, IdToken, StopWordToken. These are all just bare UserString subclasses (meaning they have `str()`, `__eq__`, `len()` etc. based on their string value, and other than that they only vary by type), except for ReporterToken. ReporterToken also tracks its exact_editions (like EDITIONS_LOOKUP["U.S."]), its variation_editions (like VARIATIONS_LOOKUP["U. S."]), and all_editions (both combined).
  *  For review: the token classes could also live in models.py.
* tokenize() emits a list of those tokens instead of a list of strings.
* get_citations() checks the types of tokens instead of inspecting the strings directly.
* When creating a `Citation`, the `exact_editions`, `variation_editions`, and `all_editions` values are copied over to the citation from the `ReporterToken`, so we can use them for disambiguation or the user can use them for their own disambiguation later.
* Disambiguating: `disambiguate_reporters` can now be split into two jobs. Right after the citation and context are extracted, we call `guess_edition` to try to narrow down multiple editions to one on *all* citations (because why not). This sets `edition_guess`. Then if the `remove_ambiguous` parameter is true, we strip out all citations where the number couldn't be reduced to one.
  * For review: in writing `models.Citation.guess_edition()` I used a shorter algorithm than the existing `disambiguate_reporters`. Pretty much just: `editions = self.exact_editions or self.variation_editions`; strip editions where the year excludes the edition; if one edition remains, that's our guess. All the tests pass with this approach, but let me know if this leaves out a nuance of the existing code.
  * For review: we no longer set `lookup_index`. I think that was an implementation detail that is no longer needed. LMK if it was serving some other purpose.
  * For review: one functional difference is I think the existing code checks dates by seeing if "1 S.W.2d 1 (1900)" falls in the date range for *any* S.W. reporter, and my new code only checks the dates for S.W.2d. That seems right to me?
* There's a bit of refactoring around is_scotus and is_neutral_tc_reporter, since those can now be booleans on the Reporter object.
* Tests:
  * I restricted the test to only checking a set list of attributes in `match_attrs`, so I wouldn't have to edit every test case with all the `*_editions` attributes for now. I can add those later if this all seems OK and we want them in.
  * Add a very simple custom tokenizer test.

## Left undone

We might want to revisit what we're exposing as far as the exact matched string, the guess about the corrected reporter string, the guess about the canonical reporter string, and the various reporter objects I'm now attaching. I didn't change any of the existing values, but maybe worth clarifying the terminology and removing redundancy once the dust settles.

### DIFF
--- a/eyecite/find_citations.py
+++ b/eyecite/find_citations.py
@@ -1,4 +1,4 @@
-from typing import Callable, Iterable, List, Optional, Sequence, Union, cast
+from typing import Callable, Iterable, List, Optional, Union, cast
 
 from eyecite.helpers import (
     add_defendant,
@@ -22,7 +22,8 @@ from eyecite.reporter_tokenizer import (
     ReporterToken,
     SectionToken,
     SupraToken,
-    Token,
+    TokenOrStr,
+    Tokens,
     tokenize,
 )
 from eyecite.utils import clean_text, strip_punct
@@ -34,7 +35,7 @@ def get_citations(
     do_defendant: bool = True,
     remove_ambiguous: bool = False,
     clean: Iterable[Union[str, Callable[[str], str]]] = ("whitespace",),
-    tokenizer: Callable[[str], Iterable[Token]] = tokenize,
+    tokenizer: Callable[[str], Iterable[TokenOrStr]] = tokenize,
 ) -> Iterable[Union[NonopinionCitation, Citation]]:
     """Main function"""
     if text == "this":
@@ -116,7 +117,7 @@ def get_citations(
 
 
 def extract_full_citation(
-    words: Sequence[Token],
+    words: Tokens,
     reporter_index: int,
 ) -> Optional[FullCitation]:
     """Given a list of words and the index of a federal reporter, look before
@@ -175,7 +176,7 @@ def extract_full_citation(
 
 
 def extract_shortform_citation(
-    words: Sequence[Token],
+    words: Tokens,
     reporter_index: int,
 ) -> Optional[ShortformCitation]:
     """Given a list of words and the index of a federal reporter, look before
@@ -237,7 +238,7 @@ def extract_shortform_citation(
 
 
 def extract_supra_citation(
-    words: Sequence[Token],
+    words: Tokens,
     supra_index: int,
 ) -> Optional[SupraCitation]:
     """Given a list of words and the index of a supra token, look before
@@ -275,7 +276,7 @@ def extract_supra_citation(
 
 
 def extract_id_citation(
-    words: Sequence[Token],
+    words: Tokens,
     id_index: int,
 ) -> Optional[IdCitation]:
     """Given a list of words and the index of an id token, gather the
--- a/eyecite/find_citations.py
+++ b/eyecite/find_citations.py
@@ -1,13 +1,10 @@
-import re
-from typing import Callable, Iterable, List, Optional, Union
+from typing import Callable, Iterable, List, Optional, Sequence, Union, cast
 
 from eyecite.helpers import (
-    REPORTER_STRINGS,
     add_defendant,
     add_post_citation,
     disambiguate_reporters,
     is_neutral_tc_reporter,
-    is_scotus_reporter,
     joke_cite,
     parse_page,
     remove_address_citations,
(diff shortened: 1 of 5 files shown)


jcushman:
Oh, I think moving to object tokens instead of string tokens may help with some other stuff later as well -- for example if we want to expose the exact string match "1 U. S., 1" that becomes easy to do by attaching start and end offsets to each token and exposing `text[volume_token.start:page_token:end]`.

jcushman:
Oh, this is also a key step along the way to handling atypical cite formats. Once we're emitting tokens, we can emit CitationTokens for `2020-NMSC-01` and tax court citations and so on.

mlissner:
I've briefly looked at this, and will give it a proper look early next week, but one thing I didn't see is a performance analysis. Any thoughts or experiments on that?

jcushman:
Good question, I hadn't run a profile yet.

The short answer is it doesn't matter much either way. 80% of runtime goes to `strings = REPORTER_RE.split(text)` in both cases, so it's lost in the noise. If we pretend we've fixed that line with hyperscan or something, then the PR was (very roughly) 10 or 20% slower on the remaining parts. I added a couple of performance tweaks that make it (very roughly) 10 or 20% faster instead. The performance fix was for non-special tokens to be `str` instead of `Token` instances, which all the code was already happy with since Token and str basically act the same anyway.

jcushman:
(Something is weird with tests but I think they should be OK.)

jcushman:
I'm not far from having a next iteration that substantially refactors this to do extraction via regex that includes page and volume numbers, so it may make sense to wait to review that one -- though that'll be a huge PR so this one might be helpful to see as evolution.

mlissner:
Thanks for the warning, Jack. I'll hang tight on reviewing until you tell me to take another look.

mlissner:
Looks like tests aren't happy. Should I continue holding off on review here?

jcushman:
Yeah, I'll send a larger thing today.

PR #12, opened 2021-01-29
Extract citations with regular expressions

jcushman:
OK, this is a big one. :) This PR updates the tokenizer so it extracts entire volume-reporter-page citations using one regex per reporter string, rather than splitting on reporter names and assembling citations from the individual tokens.

There's more work to do here, but I believe the correct foundations are here and it's now at the point where it works better than the existing thing, so I'd love to get this in as a baseline we can improve on.

## Motivation

This new design has a few benefits:

* Handling unusual formats -- "T.C. Memo 2019-1234", "2020-NMSC-01", et al. Instead of requiring custom code for each edge case, these can now be handled correctly just by setting the correct value in reporters-db `"cite_format"`.
* Related to the above, narrowing the impact of special cases. Want to handle citations to volume "25 Supp." of the Texas reports, or weird page numbers in Connecticut? You can now just tweak the regex for the Texas reporters to include `|25 Supp\.` instead of complicating parsing for every reporter.
* Supporting non-caselaw citations. I didn't take this on yet, but extracting statutory cites will take very little code once this is in place, and will be mostly about making the configuration file.

## General approach

Before getting into the code, here's broadly how we want this to work: we're adding the concept of an "Extractor," which runs a given regex on the target text and spits out objects of a given class. So the US Reports extractor, for example, has the job of running a `(?P<volume>...) (?P<reporter>U\.S\.) (?P<page>)` regex. Run on "foo v. bar, 123 U.S. 456", it will tell us that the portion of the string from offsets 12 to 24 maps to the object `CitationToken(volume="123", reporter="U.S.", page="456", exact_editions=[us_edition])`. We can define extractors for long cites, short cites, id. cites, supra cites, statutory cites, whatever we want -- they each just slice out matching parts of the target text and turn them into the appropriate objects. 

Strings that aren't matched by any extractor get split on whitespace just like before, so we end up with `["foo", StopToken("v"), "bar,", CitationToken(...)]`.

This means the hard part becomes preparing our list of Extractors from reporters-db and running them all efficiently. Once tokenizing is done, we're mostly home.

## Performance

Running a thousand regular expressions that include volume and page numbers on a large text is unworkably slow, even if we stick them together with `|` and do it in one pass.

This PR includes two approaches to handle that. The first uses the [Aho-Corasick algorithm](https://en.wikipedia.org/wiki/Aho%E2%80%93Corasick_algorithm) as implemented by [pyahocorasick](https://pypi.org/project/pyahocorasick/) to filter down our Extractors to only those that can possibly match. Each Extractor provides a list of strings that must exist for the regex to match; for the US extractor that would be "U.S." Aho-Corasick tells us which of our extractor strings exist in the target text, and we run just those ones. This is the approach used by default.

The second approach uses [hyperscan](https://pypi.org/project/hyperscan/), which is basically Aho-Corasick generalized to regular expressions. Hyperscan compiles thousands of regexes into a single universal regex that can find all matches in the target text in a single pass. This is not the default, because hyperscan is an x86 library developed by Intel, so it's only the best choice if you have an Intel or AMD chip to run it on. (Though [cross-arch forks](https://github.com/VectorCamp/vectorscan) are in progress.) Hyperscan also has a slow (several seconds) compilation step, so there's a little extra business to provide a cache directory to avoid re-running that.

Despite the issues with depending on a library developed by Intel, I believe this is a good direction to go in -- this is simply the right data structure for doing what we need to do, and it doesn't exist in any better form.

Numerically, here's a rough sense of the time it takes to run `get_citations()` on a 50 page SCOTUS case:

* this PR with hyperscan: 3ms
* current master branch (after the earlier performance optimizations): 20ms
* this PR with Aho-Corasick: 50ms
* current approach (prior to earlier performance optimizations): 170ms
* this PR with neither: 250ms

The hyperscan performance is likely to be flat (ish) as we add regexes for statutes or whatever, while the others will scale linearly (ish). I also confirmed the hyperscan performance holds up on AMD processors as well as Intel.

## Detailed notes

* Most of the notes from #10, my earlier PR, still apply. In particular the API is changed a little from master:
  * `get_citations` now accepts a `tokenizer` argument that defaults to `AhocorasickTokenizer()`.
  * `get_citations` now accepts `remove_ambiguous` (default False) instead of `disambiguate` (default True). All citations come back with `exact_editions`, `variation_editions`, and `all_editions` if you want to do your own disambiguation.
* Page number checking and punctuation stripping are handled in the regex instead of later. I think we could tighten these up a little down the line by limiting some of the special cases to particular reporters. There's also room for making the regexes overinclusive and filtering out citation matches later, but it doesn't seem to be necessary.
* Volume numbers become strings instead of int -- valid volumes aren't all numbers, alas.
* Some citations like NMCERT require updates to reporters-db to extract correctly. I tweaked the test harness a little to allow particular tests to fail on purpose, like this:
  ```
            ('2007-NMCERT-008',
             [FullCitation(volume='2007', reporter='NMCERT', page='008',
                           canonical_reporter='NMCERT',
                           reporter_index=0, reporter_found='NMCERT')],
             {'expect_fail': 'reporters.json needs NMCERT pattern'}),
  ```
  So we can easily add test cases for things that aren't working yet.

Probably lots of other stuff worthy of comment, but maybe it's easier if you flag anything in the code you'd like more comments on.

## Extraction changes from current code

The tests pass, and on my 50-page case this new code extracted most of the 70+ citations the same. There are a few differences that are improvements, because of edge cases in the old short-cite detection code:

* `"Widdicombe v. Bosemiller, 118 Fed. Rep. 295."` is no longer detected as "118 Fed., at 295"
* `"The Anna (1805), 5 C. Rob. 373."` is no longer detected as "5 C., at 373"
* `"Bice v. Minn. & N. W. B. B. Co., 1 Black, 358"` is no longer detected as "1 Black, at 358"

A change in the court-detection code is also causing nominative cites like `4 How. 638` to correctly come through as court == scotus.

I didn't note any regressions, but there may well be some. Once this lands I'd like to run it on a bunch of cases, look for what it gets wrong and add test cases.

## Left undone

Upstream changes to reporters-db:

This PR takes advantage of `"cite_format"` in reporters-db, so it can already correctly extract cites like `"T.C. Memo 2019-1234"` without any custom code. Some reporters don't extract correctly yet but will when `"cite_format"` is updated, like `NMCERT`.

However, it doesn't yet take advantage of the `"regexes"` field of reporters-db. I can get that working, but one reason I paused is that the reporters-db format doesn't provide a way to map regexes to editions. I think "regexes" wants to be a dict like "variations" that maps to "editions" keys? (OTOH currently there are no examples with regexes and multiple editions.)

There's also no elegant way to specify variants with their own cite_formats. For example, the current tests suggest that the `UNITED STATES TAX COURT REPORT` variant has parentheses around the page number. That could be handled with a regex, but it would be neat if there was some way to express `"UNITED STATES TAX COURT REPORT" -> "{volume} {reporter} ({page})"`.

API:

I suspect that, with this stuff in place, it makes sense to take another look at what's returned from get_citations() -- do the fields on Citation objects still make sense? But that's for later.


mlissner:
Whew. I had to do some research to understand just the description of this one. That's done, so now I'll turn to the code. Alas, it's Friday and my day is nearly spent, but I'll continue reading. 

Two small questions so far:

1. Why do those citations get parsed incorrectly currently? I'm surprised to see that they're wrong and that this fixes them.

2. Would it be easier to update reporters_db and tweak the requirements here to point to that branch, instead of the tweaks to the test harness?

mlissner:
This is quite a PR.

jcushman:
> Why do those citations get parsed incorrectly currently?

You mean the stuff like this (master branch)?

```
In [2]: get_citations("Widdicombe v. Bosemiller, 118 Fed. Rep. 295.")
Out[2]: [Bosemiller,, 118 F., at 295]
```

Basically "<number> <reporter> <any token> <number>" reads as a short cite. Taking that example -- reporters-db currently has "Fed." but not "Fed. Rep." (we should add "Fed. Rep.", but ignore that for a minute). So the current code finds "Fed." as the reporter. Then `extract_full_citation` fails, because "Rep." isn't a page number, so it goes to `extract_shortform_citation`, which assumes that "Rep." is a comma, tries the next token and claims "295" as the page number.

The new code just looks for `(?P<volume>...) (?P<reporter>Fed\.),? (?<page>...)` as the long-form citation and `(?P<volume>...) (?P<reporter>Fed\.),? at (?<page>...)` as the short-form, so it finds nothing, which is correct in this instance since "Fed. Rep." isn't a reporter as far as we know.

> Would it be easier to update reporters_db and tweak the requirements here to point to that branch, instead of the tweaks to the test harness?

I think the two repos shouldn't "know" anything about each other beyond what's on pypi, right? Like this one should just pull in the latest released version for tests.

As far as the changes to the test harness, this is just a gut instinct but I like the idea of being able to write the test for a failure case separately from fixing the problem, especially when the fix might have to be coordinated with a separate repo. I added this feature when I was getting ready to comment out some failing test cases and felt like I was just adding dead weight that would be easy to lose track of; keeping the failing ones noisily in felt more complete. I expect we'll encounter endless bugs that are more or less confounding to fix, and there might be less bookkeeping all around if the process when you discover an extraction bug is "write the failing test case, commit that, then just mark it as a success case when you get around to fixing it or it fixes itself with a reporters-db update."

> This is quite a PR.

😆 I would have sent less if I could but I couldn't figure out how to make it a complete thought.

jcushman:
Another "Left undone" thing here is how to insert links to the extracted citations in the source text. I haven't looked at how CL does this, but I'm assuming it uses the `as_regex()` methods, which assume "<volume> <reporter> <page>" format. I'm assuming that already doesn't work for e.g. "T.C. Memo <volume>-<reporter>" cites, and those need special handling? This PR doesn't make that worse but will return more cites where it's a problem.

The approach I'm thinking of is to have each returned `Citation` include the exact matched text, like `Citation(source_text="123 U.S. 456")` or "T.C. Memo. 1234-56", so callers can wrap the text with whatever they want. And maybe offsets as well, though those are less helpful if you're using cleaning functions. Should be a small step from what's here since we already have the match object.

mlissner:
@mattdahl do you have thoughts about Jack's last comment? I've honestly lost track of how the regex replacements work. 

----

The rest of your comments make sense, thanks Jack. What I was thinking about the reporters_db was that you could fix it too, then we could either do a release (not a big deal), or we could just reference the commit hash in this repo's branch until that release is out. The current approach is fine though.

mlissner:
OK, I've read through this. @mattdahl if you have any time to take a look at it too, I fully expect you'll have better feedback than I will. Either way, I'll try to finish this review tomorrow.

eyecite/find_citations.py
```diff
@@ -116,9 +106,9 @@ def get_citations(
 
 
 def extract_full_citation(
-    words: List[str],
+    words: Tokens,
     reporter_index: int,
```
mattdahl, line 110: I feel like this should be renamed (also in the other `extract` functions), since this is no longer the index of the reporter itself (but also the volume, page, etc.). Maybe `token_index`? (Though that is a bit vague since `words` is a sequences of `Tokens`.) `matched_token_index`?

mattdahl:
Jack's last comment makes sense to me, and that approach would probably simplify our regex generation. However, even if we used the full original matched source text as the starting point for the regex string, we would still have to deal with problem of ensuring that the regex properly matches whitespace and intervening HTML characters. The root of the issue is that the "original" source text that `get_citations()` consumes is artificially sanitized (via the `clean` parameter), so we can't actually rely on that sanitized version of the text to be the text that really appears in the original document (which might contain weird whitespace, line breaks in the middle of the citation, HTML markup, etc.). So we would still have to be careful about constructing a regex string that accounts for those annoyances. But I think that having something like a `Citation.source_text` field would make sense and would simplify things somewhat. It would also provide an entry point into fixing the tax court problem, which I believe you are correct in saying is not currently handled at all.

I don't really have time to do a full review of these changes, except to say that they look generally amazing. The basic intuition -- capture more data at the tokenizing stage rather than at the scanning stage -- seems great and obviates a lot of brittle code.

I just made one comment, above, that stuck out to me as I was reading through!

jcushman:
> we would still have to deal with problem of ensuring that the regex properly matches whitespace and intervening HTML characters

This makes me laugh guiltily every time I think about it, but one approach that works fine if your HTML tags can be matched by regex (ours can, cause our HTML is self-generated and pretty regular) is to insert a tag-matching regex between each character of the target string:

```
In [4]: import re
   ...:
   ...: def add_link(text, link_text, link_url):
   ...:     regex = r')(<[^>]+>)?('.join(r'\s+' if c == ' ' else re.escape(c) for c in link_text)
   ...:     link = f'<a href="{link_url}">'
   ...:
   ...:     def replacement(m):
   ...:         return link + ''.join(
   ...:             '' if not x else x if len(x) == 1 else f'</a>{x}{link}'
   ...:             for x in m.groups()
   ...:         ) + '</a>'
   ...:
   ...:     return re.sub(rf'\b({regex})\b', replacement, text)
   ...:

In [5]: add_link("foo <i>bar</i>\tbaz boo", "bar baz", "http://example.com")
Out[5]: 'foo <i><a href="http://example.com">bar</a></i><a href="http://example.com">\tbaz</a> boo'
```

It's evil, but ... fine?

Beyond that, eyecite could diff the text before and after cleaning to come up with a mapping of offsets, so it can tell you that offsets 12-24 in the cleaned text correspond to offsets [14-18, 20-28] in the original text and you can insert your links at those offsets. I've done something like this for annotations in other contexts and it works OK. The diff step to get that data structure is a little tricky to get right but then it's a nice generic bulletproof solution.

jcushman:
Looks like the diff algorithm I mentioned above is implemented in the diff-match-patch library already, so we should be able to provide a very short helper function that transfers links from the clean text to the original text. I can send a PR to demonstrate that once this lands.

eyecite/find_citations.py
jcushman: Good point -- I just switched them all to `index`, as well as updating the docstrings. I didn't change `reporter_index` on the `Citation` object returned by those extract functions -- as I mentioned before I think we'll want to rethink what's exposed on the `Citation` object in general, so easier to fix it then.

mlissner:
Oooook, not gonna lie, this is one of the bigger and more complex PRs I've ever reviewed. In the end, I pretty much get it, but I don't know the two new libraries here, and I haven't used a number of the new python features either. So...lots to learn, but learn I have — mostly. 

Where this stands I have no objection to merging this. I wanted to see some specific things in the tests — and I did. I wanted to read through the code and get a sense that all the old features are more or less in place — and I did. I wanted to look for stylistic things that could be more clear, or that could be improved — and I did. 

I found one thing that could probably be tweaked to be better. In the tokenizer, you have a big repeated block for filtering out overlapping matches. Seems like a candidate for factoring out, since it only differs by one or two lines. 

The rest...looks cool. Once this is merged, I suppose we'll want to do some tweaks to the reporters DB, then we can probably release this as 2.0, right?

mlissner:
Also this: https://twitter.com/FreeLawProject/status/1357492125693222914

Thank you both for how cool this all is. I couldn't find either of you on Twitter, so didn't mention @-you, but thank you.

jcushman:
> Oooook, not gonna lie, this is one of the bigger and more complex PRs I've ever reviewed.

To be fair, "here's an unsolicited +1200, -600 line change to replace your tokenizer with regular expressions" is by far the most aggro PR I've ever sent. :)

> I found one thing that could probably be tweaked to be better. In the tokenizer, you have a big repeated block for filtering out overlapping matches. Seems like a candidate for factoring out, since it only differs by one or two lines.

Good call, that was a pain point but I couldn't work out how to abstract it since one was dealing with strings and the other with bytes. I think I found a clean solution though, so we're hopefully ready to go.

> Once this is merged, I suppose we'll want to do some tweaks to the reporters DB, then we can probably release this as 2.0, right?

Yeah I think other than what's currently on master for reporters-db, what's left over there is to add regexes for tax court so we can match "1 UNITED STATES TAX COURT REPORT (2018)", and do a release so eyecite sees the updates.

Over here, these two things might be worth it before v2:

* Use regexes from reporters-db
* Add functions for inserting links into pre-cleaned text

Feel free to release whenever, but I expect I can send PRs for those things shortly.

mlissner:
OK, great. Merged. I'll wait until the last bits and pieces are landed to do a release, unless somebody wants otherwise.

PR #13, opened 2021-02-08
New annotate function for inserting links to citations

jcushman:
With apologies for another big PR, this one takes a whack at link insertion (which I'm also calling "annotation"), including the problem of inserting citation links into text containing HTML or other stuff that has been cleaned.

The upshot of this PR is we can now do this (example from the updated README):

```
    from eyecite import get_citations, annotate, clean_text

    text = '<p>bob lissner v. <i>test   1 U.S.</i> 12,   347-348 (4th Cir. 1982)</p>'
    cleaned_text = clean_text(text, ['html', 'whitespace'])
    citations = get_citations(cleaned_text)
    linked_text = annotate(cleaned_text, [[c.span(), "<a>", "</a>"] for c in citations], target_text=text)

    returns:
    '<p>bob lissner v. <i>test   <a>1 U.S.</a></i><a> 12</a>,   347-348 (4th Cir. 1982)</body>'
```

Note that it handles wrapping an `<i>` tag that overlaps the citation.

This replaces the existing `as_regex` mechanism for inserting links, and has a few benefits: 

* Can insert into text containing HTML (or into text cleaned by other arbitrary cleaning functions)
* Can insert different links for different Id. tags, where previously all "Id. at 3" cites would have the same as_regex.
* Can insert links to non-standard citation formats like NMSC and tax court

The game plan for the PR is:

1. Track the character offsets (relative to the cleaned text) that each token came from
2. Track the token that each citation came from
3. Use the offsets to insert annotations into the cleaned text
4. Use difflib to shift those offsets to where they need to be in the original text

This got big mostly because the first step of tracking offsets for each citation took some refactoring.

### Models refactor: new citation class hierarchy, dataclasses, token references

* Refactor the `Citation` objects so that all citation objects inherit from `BaseCitation`, and all citation objects know the `token` and `index` they came from. This makes it easy to track the original text and offsets that each citation came from.
* Use dataclasses for the class hierarchy, with `eq=True` and `frozen=True` (if immutable) or `unsafe_hash=True` (if mutable) to allow for hash comparisons. This just saves some boilerplate.
* Rename `FullCitation` to `FullCaseCitation` and `ShortformCitation` to `ShortCaseCitation`, for future statute extraction

### Add test factories

Changing the citation object meant changing every line of the test cases, which was a drag. I added `case_citation`, `supra_citation`, etc. factories to a tests/factories.py file so each test line would take less boilerplate and be easier to change in the future.

### Track token offsets (start and end)

Each `Token` now knows its `start` and `end` offset. This required a little messing around with the HyperscanTokenizer to translate from byte to string offsets. I also moved a couple of tests from TokenizeTest to FindTest because they weren't specific to tokenizing and it was annoying to get the offsets set up.

### Add annotate function

Now that each extracted citation knows its start and end offsets in the cleaned text, we can add the actual annotation function to annotate.py. This is pretty straightforward (build up a list with the cleaned text interspersed with the annotations at the given offsets), except for the part of shifting offsets from cleaned text to original text using the `SpanUpdater`. That part is kind of hairy -- we use a diff to generate a mapping of each range of characters in the cleaned text to a range of characters in the original text.

I played with adding a dependency to the diff-match-patch library, which produces minimal edit sequences and which I've had to use in other cases when messing with moving annotations, but it didn't seem to matter in this case so I didn't add the dependency. I kept a comment in showing how one would use that library though, in case we see weirdness when running this on more cases with the default difflib.

### Default cleaners to none

If you're going to use `annotate` you can't let `get_citations()` do cleaning itself, so I changed the default from `get_citations(clean=['whitespace'])` to `get_citations(clean=[])`.

### Expose public functions at base package for import

Since I was adding another public method I took this opportunity to update __init__.py so you can do `from eyecite import get_citations, annotate, clean_text`.

### Remove as_regex methods

I deleted the `as_regex` methods -- I think these are no longer helpful as they don't work for non-standard case citations or non-unique Id. and Supra citations. Note that their behavior may be a little different as some of them tried to capture more parts of the citation, and `CitationBase.span()` only covers the actual matched token.

### Allow extraction of Id. cites at end of text

In writing tests for the annotation function I realized that eyecite can't handle "Id." at the very end of the text (the after_tokens scanning code throws an index error), so this fixes that.

### Update readme for model refactors and annotation

Update the readme for all the changes above.


mlissner:
I started reviewing the code itself, but before I got too far along, I got confused about the API and what it's accomplishing. The crux of my confusion was why we were doing "cleaning" work, which made me think that the results of the cleaning were permanent for some reason. It couldn't be that we were doing cleaning text as part of the citation stuff could it? 

Eventually, I realized/remembered that the reason we "clean" the text is that finding citations in HTML with whitespace, and underscores and whatever else breaks regexes, so if we want to find citations, that junk has to get out of the way somehow. Got it. 

I wonder if we need a different word here. Maybe call these strippers, and maybe tweak the parameter names for get_citations so that it takes plain_text instead of just "text"? I think the docs could use a tweak too, to explain why these are separate steps so that I can remember this stuff in the future. 

Does this resonate with you guys?

mlissner:
Oh, also, I wonder if there's any point in continuing to provide a `clean` parameter to the get_citations function. Seems like a separation there might be better. `get_citations` always takes plaintext, full stop. If you have HTML, you better use the cleaning functions *first* and *always*.

jcushman:
Good ideas. I removed the `clean` option from `get_citations`, and added a section to the readme to explain what that function is expecting as far as clean input text.

Not sure about alternative terminology -- maybe "prepare_text"? Or maybe "clean_text" works when it's documented? Calling it strip_text isn't quite right because cleaning text might include replacements, like replacing curly quotes with straight quotes.

Speaking of which, that reminded me to add a test for replacing curly quotes with straight quotes, which exposed an additional case I had to handle in `SpanUpdater`. As you can see in the new test, annotating text with a replacement ends up looking like this:

`<a>1 Abbott</a>’<a>s Pr.Rep. 1</a>`

-- the curly quote gets wrapped. I think this behavior is correct, and the _most_ correct thing to do as a consumer of the library is to call annotate with `wrap_elisions=False` and custom tags as start and end markers, then use the custom tags to insert links using an HTML-aware annotation library. But this is a good place to start.

jcushman:
(I fixed the lint test failure -- pylint was getting a lower score on CI than in local dev because it needed the hyperscan library installed in order to lint those lines properly.)

A related naming-improvement question to `clean_text` might be the term `target_text` -- I struggled to name that one. That is, how to have an intuitive API for these two calls:

* Apply annotations directly to `text`, from which the citations were extracted: `annotate(text, annotations)`
* Apply annotations to `text`, given citations that were extracted from `cleaned_text`: `annotate(cleaned_text, annotations, target_text=text)`

Maybe there's a clearer interface to put on that than "you always supply the clean text the citations actually came from, and if you want to annotate different text that you got the clean text from, you supply that as an optional parameter `target_text`"? It's not the worst but maybe there's something clearer.

mlissner:
Yeah, I played with the naming a bit. What about drawing a distinction between plain_text and source_text. So you wind up with:

```
plain_text = clean_text(source_text)
citations = get_citations(plain_text=plain_text))
annoted_text = annotate(plain_text=plain_text, annotations=annotations, source_text=source_text)
```

With the idea being that we make it clear where we expect plain text, and we use `source_text` to indicate which text we're talking about. I think `target_text` might be confusing because it refers to an implementation detail. That is, the value passed there is the target of the annotations, but  as somebody that doesn't know what's going on inside the function, I don't think of it that way. I think of it as my starting text that I was working with before I had to clean it up to do citation stuff.

mlissner:
I was experimenting with the change, so I just went ahead and pushed it. But it's one commit, so we can revert if we find something we like better or if you don't like it.

mlissner:
OK, I've reviewed everything, and it all looks good. @mattdahl said he'd take a look, probably this weekend, with an eye towards whether any of this would be tricky on the CL side of things. I don't think it would, but he knows that part better. Matt, if you just want to look at the API of this and think about whether it's sufficient, that might be all we need from your end. 

Jack, if Matt says it's good, I think we just have the parameter naming question to wrestle with. The rest looks solid to me.

jcushman:
I don't feel strongly about the naming on this one -- your tweak is fine with me!

mlissner:
Awesome, yeah, I didn't feel strongly about them either. I think my tweaks are an improvement, but I still don't quite love the API. Maybe it's fine though and I just need to get used to it.

With that, we're just waiting for your review then, @mattdahl. I think it's ready to go, but I wouldn't hesitate to make changes if you noted some we needed for this tweak. 

Thank you both!

README.rst
```diff
@@ -19,59 +19,117 @@ Simply feed in a raw string of text (or HTML), and receive a list of structured
...
+Note that the annotations are wrapped around each spot where the cleaning steps removed text
+from the citation: "<a>1 U.S.</a></i><a> 12</a>". This ensures that :code:`annotate` will
+emit valid HTML. If you don't want that, use :code:`wrap_elisions=False`:
+
+::
+
+    linked_text = annotate(plain_text, [[c, "<a>", "</a>"] for c in citations], source_text=source_text, wrap_elisions=False)
+
+    returns:
+    '<p>bob lissner v. <i>test   <a>1 U.S.</i> 12</a>,   347-348 (4th Cir. 1982)</p>'
```
mattdahl, line 132: IMO it would be preferable for the default behavior to be to _not_ wrap elisions (i.e., to annotate the entire string as one). But, as you point out, the risk of doing this is that sometimes broken HTML will be emitted (if the original text has intervening HTML tags, as in this example). To overcome this problem in CL I created a function `is_balanced_html()` (https://github.com/freelawproject/courtlistener/blob/master/cl/citations/utils.py#L47) so we could first check whether the text-to-be-annotated indeed contained such intervening tags (i.e., was "unbalanced" HTML). Could we do something similar here?

This would be my ideal behavior, I think:
- `annotate()` tries to wrap the entire string as one (no flag needed)
- If `annotate()` cannot do so safely (i.e., the text-to-be-annotated fails the `is_balanced_html()` check), then a new flag determines how it proceeds, in one of three ways:
  - It does the full wrap anyway, knowing that broken HTML will be emitted
  - It safely wraps around the individual elisions instead (the current behavior under `wrap_elisions=True`)
  - It does no annotation at all

Thoughts?

eyecite/annotate.py
```diff
@@ -0,0 +1,174 @@
...
+            elif operation == "replace":
+                yield "delete", a2 - a1
+                yield "insert", b2 - b1
+            else:  # 'delete', 'equal'
+                yield operation, a2 - a1
+
+    # same as above using the diff-match-patch library, which may
+    # work better with some inputs because it offers minimal edit sequences:
+    # @staticmethod
+    # def get_dmp_diff_steps(self, a: str, b: str):
```
mattdahl, line 145: Why is this commented out? Should you expose the option of using this method instead to the user?

eyecite/models.py
```diff
@@ -41,75 +44,65 @@ def includes_year(
...
+    token: "Token"  # token this citation came from
+    index: int  # index of _token in the token list
+
+    def matched_text(self):
+        """Text that identified this citation, such as '1 U.S. 1' or 'Id.'"""
+        return str(self.token)
+
+    def span(self):
+        """Start and stop offsets in source text for matched_text.()"""
+        return (self.token.start, self.token.end)
```
mattdahl, line 60: Can the `IdCitation` class override this method to somehow extend the calculated `end` value to include the end of its `after_tokens`, if its `has_page == True`? (Idea being that something like "Id., at 123" should be annotated fully, not just the "Id." part.)

mattdahl: Same thing with `SupraCitation` actually, with its `antecedent_guess` and possible `page`?

README.rst
jcushman: I'll take a stab at this -- it's nice to not wrap text replacements like smart quotes. I think the default wants to be no check, since "this is HTML with spans and I should do an expensive parsing step for each string" is something you should have to opt into.

eyecite/annotate.py
jcushman: I can do that. My hunch is there's one right answer, and we'll find out which one shortly after we run this at scale. But I'm fine with having an option while we sort it out.

eyecite/models.py
jcushman: Huh, yeah. Id with has_page=True makes sense to me; antecedent_guess strikes me as too noisy. I can poke at it.

jcushman: I should also double check that annotate can handle overlapping annotations.

README.rst
mlissner: OTOH, I think if it's the only way to get good results, that kind of expensive parsing might just be the cost of doing business, and having defaults with good (albeit slow) results might just be OK.

jcushman: My latest update just inserts annotations at the edges of the citation by default, but accepts `unbalanced_tags="wrap"` to wrap HTML tags in unbalanced tags are detected, and `unbalanced_tags="skip"` to skip those annotations.

One choice I made here is to only wrap HTML tags, using a new `wrap_html_tags` helper, and not wrap all elisions like extra spaces. This is prettier, and also I thought it would be a weird behavior if spaces were wrapped or not based on whether or not the citation also contained an unbalanced tag.

A second choice I made was to assume straightforward HTML tags in `wrap_html_tags` that can be matched with `r"<[^>]+>"`. This isn't correct for all HTML, but wrapping HTML text spans is a large problem on arbitrary HTML, and I documented that as out of scope in the readme -- we're assuming your HTML is properly escaped (you could process it through lxml.html if not), and that your citations don't include anything super weird like the start of an HTML comment or a pre tag (I think this would only happen with deliberately broken input). If malicious input is a concern one should just insert self-closing markers and do the highlighting with a heavier library, maybe in javascript. (Note this choice doesn't depend on the previous choice -- you could still make broken HTML from malicious input if we were wrapping all elisions instead of just tags.)

> I think if it's the only way to get good results, that kind of expensive parsing might just be the cost of doing business, and having defaults with good (albeit slow) results might just be OK.

For this version I stuck with the default of not checking for unbalanced tags, though it's not a huge deal to me either way. I agree that wrapping is a good default _if_ your source_text is (a) HTML (b) that is reasonably well-formed but (c) has potential markup in the citations. I just think that's specific enough that you should have to think about it and make sure it applies to you; it shouldn't be an implicit contract of the default `annotate()` call.

This isn't a performance issue since we can short-circuit most strings by checking for angle brackets, it's more just a heavy amount of code to run that might do stuff that doesn't make sense outside our particular situation.

eyecite/annotate.py
jcushman: Added a (for-now undocumented) option `annotate(use_dmp=True)`, defaulting to False. Using this option requires `pip install diff-match-patch`.

eyecite/models.py
jcushman: Added handling for IdCitation, and a test showing `<0>1 U.S. 1</0>. Foo. <1>Id.</1> Bar. <2>Id. at 2.</2>`. 

I didn't take on SupraCitation, which currently looks both over-inclusive and under-inclusive to me in setting `page` and `antecedent_guess`. I'd suggest saving that for a separate pass that includes seeing if those need to be tweaked based on what they're currently matching.

One implementation detail -- we have to manually calculate the end offset in `IdCitation.span` by adding up the length of `after_tokens` strings, because regular strings aren't `Token` objects. It might be worth taking a performance hit to have everything be a token with offsets tracked rather than do that fussy math. I'd also like to save that for a future look though.

mattdahl: Yeah, I agree that SupraCitation is problematic, probably best to leave that for latter.

mattdahl:
Thanks for making those changes, @jcushman. This looks really amazing! I have no objection to it being merged @mlissner.

mlissner:
Seriously sorry about letting this one slip. I was deep into setting up our intrusion detection system last week. This looks great and I absolutely love it. The next steps, at this point, I presume are to actually integrate this, right? Or did we have more tweaks to the reporters DB that were needed?

PR #31, opened 2021-03-03
Use custom regexes from reporters-db

jcushman:
This is a first pass at using the `"regexes"` field from reporters-db to extract non-standard citations. E.g. from the tests:

```
            # Test reporter with custom regex
            ('blah blah Bankr. L. Rep. (CCH) P12,345. blah blah',
             [case_citation(2, volume='', reporter='Bankr. L. Rep.',
                            reporter_found='Bankr. L. Rep. (CCH)', page='P12,345')]),
            ('blah blah, 2009 12345 (La.App. 1 Cir. 05/10/10). blah blah',
             [case_citation(2, volume='2009', reporter='La.App. 1 Cir.',
                            reporter_found='(La.App. 1 Cir.', page='12345')]),
```

As you can see, volume can sometimes now end up as an empty string. Custom regexes can also extract extra data -- in the last test example `cite.token.extra_match_groups == {"date_filed": "05/10/10"}`.

## Limitations on the file format

There's a few limitations of the existing reporters-db format where I had to make arbitrary choices:

**Multiple editions**

In this example, does the regex map to "Foo" or "Foo 2d"? Since this is unknowable, I just skip over reporters with `"regexes"` and multiple editions; currently there aren't any, but there could be as we improve things.

```
"Foo": [
        {
            ...
            "editions": {
                "Foo": {...},
                "Foo 2d": {...},
            },
            ...,
            "regexes": [
                "(?P<reporter>Foox) (?P<volume>\\d+)-(?P<page>\\d+)"
            ],
        }
    ]
```

**Replace or extend?**

When an entry has `"regexes"`, do we _only_ run those or do we also run the standard `"{volume} {reporter},? {page}"` regexes for the edition and variation strings? I chose to only run the regexes, because in the existing entries like `La.App. 1 Cir.` and `Bankr. L. Rep.` it would be wrong (or at least pointless) to run the standard regexes. But what if we do want to keep the existing ones? For example, suppose normal tax court citations were `1 T.C. 1`, but they could also be cited with a variant reporter and extra parens like `1 UNITED STATES TAX COURT REPORTS (1)`:

```
"T.C.": [
        {
            "editions": {
                "T.C.": {...}
            },
            "regexes": [
                "(?P<volume>\\d+) (?P<reporter>UNITED STATES TAX COURT REPORTS?) \((?P<page>\\d+)\)"
            ],
            "variations": {
                "T. C.": "T.C.",
                "T.C. at": "T.C.",
                "T.Ct": "T.C.",
                "T.Ct.": "T.C.",
                "UNITED STATES TAX COURT REPORT": "T.C.",
                "UNITED STATES TAX COURT REPORTS": "T.C."
            }
        }
    ],
```

Here you would want the custom regex to replace some of the standard regexes but not others, and we don't really have a way to express that ... although we could handle it by providing a redundant standard regex, like

```
            "regexes": [
                "(?P<volume>\\d+) (?P<reporter>T.C.|T. C.|T.C. at|...) (?P<page>\\d+)",
                "(?P<volume>\\d+) (?P<reporter>UNITED STATES TAX COURT REPORTS?) \((?P<page>\\d+)\)"
            ],
```

**Volume and page number regexes**

The regexes for `(?P<volume>...)` and `(?P<page>...)` have a lot of variation in reporters-db, and some stuff that looks non-standard or wrong like `[0-9]{1,}` and `((\d){1,3})+`. It's hard to tell what's intentional and what's just different ways to express "normal page number." This made me wonder if it would be useful to have a placeholder when nothing specific is intended about valid values, along the lines of `(?P<page>{PAGE_REGEX})`.

The above three issues made me wonder if it's worth revisiting the way reporters-db includes regexes, maybe adding a new field, but I don't have an immediate proposal for something that elegantly solves everything.

### DIFF
--- a/eyecite/models.py
+++ b/eyecite/models.py
@@ -285,6 +285,7 @@ class CitationToken(Token):
     exact_editions: Sequence[Edition] = field(default_factory=tuple)
     variation_editions: Sequence[Edition] = field(default_factory=tuple)
     short: bool = False
+    extra_match_groups: dict = field(default_factory=dict, compare=False)
 
     def __post_init__(self):
         """Make iterables into tuples to make sure we're hashable."""
@@ -301,11 +302,15 @@ class CitationToken(Token):
         "variation_editions" in their extra config. Pass all of that through
         to the constructor."""
         start, end = m.span(0)
+        match_groups = m.groupdict()
         return cls(
             m[0],
             start + offset,
             end + offset,
-            **m.groupdict(),
+            volume=match_groups.pop("volume", ""),
+            reporter=match_groups.pop("reporter", ""),
+            page=match_groups.pop("page", ""),
+            extra_match_groups=match_groups,
             **extra,
         )
 
--- a/eyecite/tokenizers.py
+++ b/eyecite/tokenizers.py
@@ -282,40 +282,36 @@ class AhocorasickTokenizer(Tokenizer):
     the target text contains one of the strings from
     TokenExtractor.strings."""
 
+    def __post_init__(self):
+        """Set up helpers to narrow down possible extractors."""
+        # Build a set of all extractors that don't list required strings
+        self.unfiltered_extractors = set(
+            e for e in EXTRACTORS if not e.strings
+        )
+        # Build a pyahocorasick filter for all case-sensitive extractors
+        self.case_sensitive_filter = self.make_ahocorasick_filter(
+            (s, e)
+            for e in EXTRACTORS
+            if e.strings and not e.flags & re.I
+            for s in e.strings
+        )
+        # Build a pyahocorasick filter for all case-insensitive extractors
+        self.case_insensitive_filter = self.make_ahocorasick_filter(
+            (s.lower(), e)
+            for e in EXTRACTORS
+            if e.strings and e.flags & re.I
+            for s in e.strings
+        )
+
     def get_extractors(self, text: str) -> Set[TokenExtractor]:
         """Override get_extractors() to filter out extractors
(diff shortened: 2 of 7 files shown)


mlissner:
This is great! A couple thoughts:

For the editions question, why not move the regexes key so that it's in the editions dict? Either that, or make the regexes key to a dict that matches the edition keys:

```
regexes: {F: [some_regex1, some_regex2], "F. 2d": [some_regex3]}
```
I guess that could have some redundancy since we don't have a good way to handle a regex that applies across editions, but that might not be so bad.

--------------

I think replacing rather than extending makes sense. If citations are normally standard form, but sometimes a given reporter needs the standard form *and* some special regex, then you just provide the standard form and the regex in the regexes dict. Should work nicely, right?

----------

For templatizing the regexes, we actually have a pretty good model for that in the courts DB:

https://github.com/freelawproject/courts-db/blob/master/courts_db/utils.py#L34

That DB has a TON of regexes, and they're all quite similar, so we wound up pulling out common parts into templates of a kind. We didn't want extra dependencies, so we just use built in Python tools for it.

I don't know what's up with the weird/broken regexes you found. I suspect they're just kind of drafty. @flooie might have more thoughts there.

-----------

I also noticed that there is both a cite_format and a regexes key in the reporters_db, and I admit I don't know why we have both. Seems like cite_format could and probably should go away. If nothing else, it should probably be documented. @flooie, do you remember what the story was there?

jcushman:
I updated this to use the new reporters-db regexes.

Issue #34, opened 2021-03-06
Annotating large strings is slow/incorrect


mattdahl:
Here's a gist: https://gist.github.com/mattdahl/21080e7aafd0a28c3ef0ebf9d9d13b0c

In short, when `use_dmp=True`, the annotation is inserted at the wrong location. When `use_dmp=False`, the location is correct, but it takes about 30 seconds to compute on my machine.

mlissner:
Hm, sounds like it might be two bugs. Performance of the non-dmp approach and accuracy of the dmp library.

jcushman:
Sorry, just saw this ...

I get correct results from `use_dmp=True` in your gist. Can you confirm that this doesn't work? If so maybe we'll have to check OS versions and such. 😬

```
from pathlib import Path
from eyecite import annotate, clean_text

opinion_text = Path('opinion_text.txt').read_text()
cleaned_text = clean_text(opinion_text, ["all_whitespace"])

result_dmp = annotate(
  plain_text=cleaned_text,
  annotations=[((902, 915), '~FOO~', '~BAR~')],
  source_text=opinion_text,
  use_dmp=True)

result_builtin = annotate(
  plain_text=cleaned_text,
  annotations=[((902, 915), '~FOO~', '~BAR~')],
  source_text=opinion_text,
  use_dmp=False)

print(result_dmp==result_builtin)
# True
```

(Both result in the annotation "~FOO~539\n  U. S. 306~BAR~").

I agree the performance of the builtin difflib is unacceptable, though! And the pure-python version of DMP isn't too fast either -- apparently it's [known to Google](https://github.com/google/diff-match-patch) to be hundreds of times slower than the C++ version. I can send a PR to use the faster DMP, and maybe we want that to be the default, assuming it's giving correct results.

mattdahl:
No worries, I should have tagged you. Infuriatingly, however, now I can't reproduce my original gist either.

However, I modified the test string to be even longer than before (now it's a full opinion from CL), and I think now something similar is happening. Edited gist: https://gist.github.com/mattdahl/21080e7aafd0a28c3ef0ebf9d9d13b0c

The native library is still slow but correct, but the `diff_match_patch` now just (incorrectly) puts the annotation at the beginning of the string. I also added a third option -- the new `diff_match_patch_python` library in your #43 -- which seems to be both correct and fast to me.

[...]

PR #43, opened 2021-03-22
Support diff_match_patch_python library

jcushman:
Per #34, this PR adds support for an alternate diffing library in `annotate()`. We now have three options (though probably want to converge on one):

* By default, `annotate()` uses built in Python diffing.
* If you `pip install diff_match_patch`, then `annotate(use_dmp=True)` uses [this pure-python DMP library](https://pypi.org/project/diff-match-patch/).
* If you instead `pip install diff_match_patch_python`, then `annotate(use_dmp=True)` uses [this C++ DMP library](https://pypi.org/project/diff_match_patch_python/). (Note for testing that you must `pip uninstall` the other one first -- they don't coexist.)

The speed difference is *very* significant: in my test on the text in #34, the first option took 17,000ms to call annotate(), the second option took 700ms, and the third option took 10ms. So I'd pretty much only consider the `diff_match_patch_python` performance to be acceptable.

There's an open question in #34 of whether DMP returns correct results, though I haven't been able to reproduce the error.

(FWIW there's also some [packaging weirdness](https://github.com/JoshData/diff_match_patch-python/issues/4) about `diff_match_patch_python` -- it changes the interface of Google's official library but uses the same name, so if you have one installed you can't use packages that rely on the other one.)

### DIFF
--- a/eyecite/annotate.py
+++ b/eyecite/annotate.py
@@ -3,6 +3,8 @@ from difflib import SequenceMatcher
 from functools import partial
 from typing import Iterable, Optional, Tuple
 
+import diff_match_patch
+
 from eyecite.utils import is_balanced_html, wrap_html_tags
 
 
@@ -11,7 +13,7 @@ def annotate(
     annotations: Iterable[Tuple[Tuple[int, int], str, str]],
     source_text: Optional[str] = None,
     unbalanced_tags: str = "unchecked",
-    use_dmp: bool = False,
+    use_dmp: bool = True,
 ):
     """Insert annotations into text around each citation.
         Each annotation is a tuple of an extracted citation, before text, and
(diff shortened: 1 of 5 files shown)


mlissner:
Kind of sounds like we should just go with the last one. The advantage of the first is no dependencies, but it sounds like it's so slow it's not worth keeping around. The advantage of the second is...no C++ (?), but I feel like there's no getting around that in a big project these days anyhow? 

I guess if @mattdahl can't reproduce his error in #34, maybe we just go with the third one and drop the others? It wouldn't change our output or APIs, right, so seems like a safe decision? (I also don't think many people are using eyecite yet in any case.)

jcushman:
Updated to use `diff_match_patch_python` by default, but kept the option around to use `difflib` just for debugging. Also added Matt's long case as a testcase.

mlissner:
Nice. I especially appreciate the move to using `+`, `-` and `=`. It made the whole thing make more sense to me.

mattdahl:
This is great, thank you!

Issue #32, opened 2021-03-04
Expose test factories


mattdahl:
I would like to use the handy object-creation factories that @jcushman created (https://github.com/freelawproject/eyecite/blob/master/tests/factories.py) to streamline CL's tests as well.

However, we (rightly imo) exclude the tests directory when we package eyecite for distribution (https://github.com/freelawproject/eyecite/blob/master/setup.py#L36).

I propose either:

1. Moving the `factories.py` file into the main `eyecite` directory; or
2. Putting the factory functions onto the Citation models directly as class methods (sort of a pseudo-constructor specifically for creating mock objects)

Thoughts?

mlissner:
Just exposing them seems like the easiest thing to me. From a documentation perspective, it seems pretty easy to explain that, "If you want to do tests in your own system, you can do so by importing xyz."

mlissner:
Maybe a pseudocode of the second approach would help me see the vision.

mlissner:
Fixed via #47

PR #47, opened 2021-03-25
Issue 32 expose test factories

mattdahl:
Re #32, simple change to expose the test factories in the packaged version of eyecite. Impetus is that I'm using them in CL, but I imagine others might want to use them too when writing their own tests.


mlissner:
Thanks. This looks great.

PR #48, opened 2021-03-25
Handle side-by-side tokens with overlapping regexes

### LINKED ISSUE #46: Tokenizer fails to tokenize adjacent StopWord, Id, and Supra tokens
If an "id" or "ibid" or "supra" reference is preceded by a stop word token, the former are not properly tokenized as their respective token types:

_Compare_
```
from eyecite.tokenizers import default_tokenizer
list(default_tokenizer.tokenize('see id. at 577.'))
# returns [StopWordToken(data='see', start=0, end=3, stop_word='see'), 'id.', 'at', '577.']
```

_with_
```
from eyecite.tokenizers import default_tokenizer
list(default_tokenizer.tokenize('see foo id. at 577.'))
[StopWordToken(data='see', start=0, end=3, stop_word='see'), 'foo', IdToken(data='id.', start=8, end=11), 'at', '577.']
```

The consequence of this is that those would-be citations are not extracted at all downstream.

I haven't really attempted to debug this, though I suspect it may have something to do with how the tokenizer deals with overlapping matches. (Just a guess.) @jcushman, do you have another intuition?

jcushman:
This fixes #46. As @mattdahl suggested, that bug is because the full regexes for "see" and "id." overlap (they both capture the space between them, but then ignore it by only grabbing what's in the match group). Unlike the regexes, the tokens generated from each regex have correct start and end offsets and don't overlap, so the core fix was just a one-liner to compare offsets on `extractor.get_token(m)` instead of the offsets of `m.span()` itself.

What made that tricky was that hyperscan works on bytes instead of strings, and dealing with the byte offsets was complicating everything. I realized that I could instead update the byte offsets right after they came out of hyperscan, and make it easier to work with everything downstream. Thus began refactoring.

So this has a bunch of changed lines, but it's hopefully just untangling some things to make any bugs like this easier to fix in the future.

### DIFF
--- a/eyecite/models.py
+++ b/eyecite/models.py
@@ -260,7 +260,7 @@ class Token(UserString):
     end: int
 
     @classmethod
-    def from_match(cls, m, extra, offset=0):
+    def from_match(cls, m, extra, offset=0) -> "Token":
         """Return a token object based on a regular expression match.
         This gets called by TokenExtractor. By default, just use the
         entire matched string."""
@@ -296,7 +296,7 @@ class CitationToken(Token):
         )
 
     @classmethod
-    def from_match(cls, m, extra, offset=0):
+    def from_match(cls, m, extra, offset=0) -> Token:
         """Citation regex matches have volume, reporter, and page match groups
         in their regular expressions, and "exact_editions" and
         "variation_editions" in their extra config. Pass all of that through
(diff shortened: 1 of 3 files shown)


eyecite/models.py
```diff
@@ -260,7 +260,7 @@ class Token(UserString):
     end: int
 
     @classmethod
-    def from_match(cls, m, extra, offset=0):
+    def from_match(cls, m, extra, offset=0) -> "Token":
```
mlissner, line 263: Should this be in quotes?

mlissner:
LGTM aside from the one comment.

jcushman:
Yeah cause it's not defined yet (until py 3.10):

https://stackoverflow.com/a/53450349

(Dunno if the from future thing would be a good idea in general.)

mlissner:
Ah, I didn't grok what it was doing at first. I'll trust your judgment here. Looks like a dumb corner case.

PR #51, opened 2021-03-29
Allow custom annotation formats

jcushman:
I ran into an edge case for CAP where citations can contain page numbers:

`1 U.S. <a class="page-label" ...>123</a>1`

I have a cleaning step that removes these links and their text for cite extraction, but then annotating doesn't work -- I get `<a href="/us/1/1/">1 U.S. <a class="page-label" ...>123</a>1</a>`, which is bad HTML, and `unbalanced_tags` can't help.

This PR adds an `annotator` argument to `annotate()` so I can pass in a callable with custom handling for this case -- see the edited Readme for an example.

As usual I'd welcome naming suggestions if `annotator` is confusing.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -136,6 +136,20 @@ regularly formatted documents such as case text published by courts. It may have
 unpredictable results for deliberately-constructed challenging inputs such as citations containing partial HTML
 comments or :code:`<pre>` tags.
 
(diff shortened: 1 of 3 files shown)


mlissner:
I dig it!

PR #53, opened 2021-04-01
fix(deps): Adds diff_match_patch_python to requirements.txt.

mattdahl:
Correct me if I'm wrong, but do we not need to add `diff_match_patch_python` as a dependency in `requirements.txt` as well as in `pyproject.toml`? I admit that I am still confused by how poetry works vs. pip, but I think `requirements.txt` is relied upon to build the dependency list when eyecite is packaged for pypi?


mlissner:
Yeah, I noticed this problem with the release yesterday, but got stuck figuring out the right solution. I think the better way to fix this is to *not have* a requirements.txt file and to just have the toml file, but I haven't been able to figure that out yet. I'm asking our resident poetry expert to see what he thinks.

mlissner:
I'm merging this with some changes that I pushed. I'm not sure if Github Actions will work like I want, but I'm going to get them fixed in `master`, where the actions work more predictably.

mlissner:
OK, this is working flawlessly. No more requirements.txt file, no more setup.py file. Good riddance.

PR #55, opened 2021-04-13
Avoid scanning tokens past end of input

jcushman:
This PR adds bounds checks and tests for a set of partial citation inputs that currently raise IndexError from `get_citations` -- 

```
'123 supra.'
', supra.'
', 1 U.S., at 1'
'1 U.S., at 1'
'v. Bar, 1 U.S. 1'
'1 U.S. 1 ('
```

The one functional change is that `get_citations('supra.')` now returns a `SupraCitation` with no antecedent, where before it would have returned nothing. I think it's less surprising to have `get_citations('supra.')` do the same thing as `get_citations('no cite supra.')` instead of being a special case.

### DIFF
--- a/eyecite/find_citations.py
+++ b/eyecite/find_citations.py
@@ -132,7 +132,8 @@ def extract_shortform_citation(
     Shortform 1: Adarand, 515 U.S., at 241
     Shortform 2: 515 U.S., at 241
     """
-    # Get antecedent
+    # Get antecedent -- either previous word, or previous two words if
+    # previous word is a comma
     antecedent_guess = None
     if index > 0:
         antecedent_guess = str(words[index - 1])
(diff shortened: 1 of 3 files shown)


mlissner:
Looks fine code-wise. Would it help future-us to have some comments in there? The `index` adding/subtracting is pretty confusing without something like that?

jcushman:
I added a few comments, and tweaked `extract_id_citation` to do the same thing with a little less `+ 1`. Looks like tests have a false failure and want to be rerun.

I didn't add specific comments for all the `if index > 0:`, because it ends up being ten different copies of "Bounds checks to ensure we don't scan before start or end of words" and started to feel like noise. Basically you have to never access `words[i]` without also deciding what makes sense to do if `i < 0` or `i > len(words)`, or clamping `i` so that can't happen, and the comment ended up feeling like it was restating what the code was doing rather than illuminating it.

mlissner:
I went ahead and merged this. I think our venv cache got screwy again, but there's a 95% chance the next PR will be from you, so I imagine you'll catch any failures on that one, if they're real. 

Thanks for this fix.

PR #57, opened 2021-04-18
fix(annotate): Toggles bisect method by start/end update; fixes #54.

### LINKED ISSUE #54: unbalanced_tags="skip" erroneously skips tightly wrapped, balanced HTML
For example,

```
from eyecite import annotate, clean_text, get_citations

s = 'foo <i>Ibid.</i> bar'
s_cleaned = clean_text(s, ['html', 'all_whitespace'])
annotate(
   plain_text=s_cleaned,
   annotations=[[get_citations(s_cleaned)[0].span(), 'A', 'Z']],
   source_text=s,
   unbalanced_tags='skip'
)
# returns 'foo <i>Ibid.</i> bar' with no annotation
```

There is HTML here, yes, but it does not bisect the the substring ("Ibid.") to be annotated -- thus, one would expect the annotation to stick even with ` unbalanced_tags='skip'`.

I think this has to do with the way the diff between the plain text and the source text is calculated. Once the new offsets for the citation are calculated here (https://github.com/freelawproject/eyecite/blob/master/eyecite/annotate.py#L59), eyecite (erroneously) thinks that the closing `</i>` tag is part of the substring to be annotated, (rightfully) detects it as unbalanced, and (rightfully) declines to do so.

I'm not immediately sure how to fix. My diagnosis of the problem may also not be complete.

mattdahl:
This PR is a simple fix to address the problem I identified in #54.

Before, we would always use `bisect_right()` to determine the index of the updater function to use when calculating the new offsets for an annotation. This makes sense when we are updating the _start_ offset for a string, because we want to use the updater whose position is `bisect_right() - 1` and `bisect_right()` returns the index that is one to the right of the offset index. However, this causes a kind of off-by-one error with the _end_ offset of a string, since we now want the index in the other (leftward) direction instead. Hence, I think we need to use `bisect_left()` for the end offsets, which returns the index that is one to the left of the offset index.

Apologies if this is confusing, it's difficult for me to explain the intuition. See more here: https://docs.python.org/3/library/bisect.html

@jcushman you know this code better than me, does this make sense to you? Do you think this introduces problems elsewhere?

### DIFF
--- a/eyecite/annotate.py
+++ b/eyecite/annotate.py
@@ -1,4 +1,4 @@
-from bisect import bisect_right
+from bisect import bisect_left, bisect_right
 from difflib import SequenceMatcher
 from functools import partial
 from typing import Any, Callable, Iterable, Optional, Tuple
@@ -55,8 +55,8 @@ def annotate(
     for (start, end), before, after in annotations:
         # if we're applying to source_text, update offsets
         if offset_updater:
-            start = offset_updater.update(start)
-            end = offset_updater.update(end)
+            start = offset_updater.update(start, bisect_right)
+            end = offset_updater.update(end, bisect_left)
 
         # handle overlaps
         if start < last_end:
@@ -196,8 +196,8 @@ class SpanUpdater:
             elif operation == "equal":
                 yield "=", a2 - a1
 
-    def update(self, offset):
+    def update(self, offset, bisect):
         """Shift an offset left or right."""
-        index = bisect_right(self.offsets, offset) - 1
+        index = bisect(self.offsets, offset) - 1
         updater = self.updaters[index]
         return updater(offset)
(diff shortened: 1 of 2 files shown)


jcushman:
Conceptually I think this is the right fix -- when an offset falls in the same place as an elision, you can choose to put it on one side of the elision or the other, and you get better results if you adjust both the start and end inward or both the start and end outward. Seems OK to me? I rely on the tests being good for this stuff because it's so hard to keep in your head ... so if this change does cause any problems we should add more tests. :)

mlissner:
If y'all are behind it and tests pass, who am I to think otherwise. Looks simple and clean. Thanks.

mattdahl:
> when an offset falls in the same place as an elision

Yes exactly, I didn't make this clear, but this is only an issue when the offset index == an elision index. That's why our existing tests didn't catch this.

PR #58, opened 2021-04-18
Framework for counting citations

mattdahl:
`get_citations()` returns a list of found citations, but what is the user to do with this list? Counting them (by resolving the `supra` and `id` citations to their antecedents) seems like the obvious next step.

This code is mainly inspired by how we tackle this problem in CL: https://github.com/freelawproject/courtlistener/blob/master/cl/citations/match_citations.py. The idea here is to provide the skeleton of a similar approach, which does the best matching possible without any API lookups or anything. However, the main `count_citations()` method is written in such a way that the user can also pass custom matching functions to it, if they have a thicker opinion abstraction or more complex matching logic that they want to use.

I think having this code in `eyecite` is useful, but let me know what you think before I continue. This PR is not yet mergeable (no tests).


eyecite/count.py
```diff
@@ -0,0 +1,162 @@
...
+        # opinion directly
+        elif isinstance(citation, FullCaseCitation):
+            matched_opinion = match_fullcase_citation(citation)
+
+        # If the citation is an id citation, just resolve it to the opinion
+        # that was matched immediately prior (so long as the previous match
+        # was successful)
+        elif isinstance(citation, IdCitation):
+            if was_matched:
+                matched_opinion = opinions[-1]
```
mlissner, line 142: Seems like this should get abstracted out too?

mlissner (commented):
A couple little thoughts, but I think I'd want to see some tests before I have a really good sense of this. I think it feels about right though and that it'll be a useful addition to those that want to use this tool. Seems better here than in CL too, since it seems impressively easy to abstract out.

eyecite/count.py
```diff
@@ -0,0 +1,162 @@
...
+    citation: SupraCitation, opinions: List[Opinion]
+) -> Optional[Opinion]:
+    """
+    Try to match supra citations by checking whether their antecedent_guess
+    appears in either the defendant or plaintiff field of the citation of any
+    of the previously matched opinions.
+    """
+    candidates: List[Opinion] = []
+    for o in opinions:
+        if citation.antecedent_guess:
```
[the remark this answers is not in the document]
mattdahl: Thanks, refactored this

eyecite/count.py
mattdahl: Touché

mattdahl:
Okay, I've added some tests and think that this is now ready for a real review. The logic of the tests is mostly copied from https://github.com/freelawproject/courtlistener/blob/master/cl/citations/tests.py#L270.

eyecite/count.py
```diff
@@ -0,0 +1,180 @@
...
+    return Opinion(citation)
+
+
+def count_citations(
+    citations: List[CitationBase],
+    match_id_citation: Callable[
+        [IdCitation, List[Any], bool], Optional[OpinionBase]
+    ] = _match_id_citation,
+    match_supra_citation: Callable[
+        [SupraCitation, List[Any]], Optional[OpinionBase]
```
mattdahl, line 108: For all of these, the call signature that I really want is `Callable[[SupraCitation, List[OpinionBase]], Optional[OpinionBase]]`, i.e., I want to specify a function that takes a supra citation (or whatever) and a list of opinions as arguments, and then optionally returns an opinion. In theory, this would then allow the user to use any custom opinion object that inherits from `OpinionBase` here.

This works fine for the _return_ type -- eyecite's own matching functions that I have written (which all return `Opinion`, a subtype of `OpinionBase`) all work without a problem.

However, this doesn't work for the _argument_ types. This is because arguments in `Callable` are apparently treated contravariantly in Python, not covariantly. See https://www.python.org/dev/peps/pep-0483/#covariance-and-contravariance for a discussion of this. Thus, the expected type inheritance is "reversed", and objects that are subtypes of `OpinionBase` are NOT accepted as arguments in a `Callable` so specified.

This seems wrong to me, but who I am to question Python's type design? In any case, my workaround is to write `List[Any]` instead of the more narrow (and more correct, imo) `List[OpinionBase]` in the `Callable` here.

jcushman:
I'm excited to see this in progress! This is a shortcoming over on the CAP side and this looks really useful.

As far as the API I feel like there might be a step before counting citations of just linking each short cite to its full cite with an attribute? Like imagine a task we wanted to do with this API was annotate each Id. cite with the URL of the full cite. Right now I think that's challenging, because you're finding but not storing the target full cite. But if the API was along the lines of ...

```
cites = get_citations()
link_short_cites(cites, ...)  # Set target_cite: Optional[FullCaseCitation] attribute of each short cite, using similar logic to the current `count_citations` function
```

... then counting can be done in a few lines based on `target_cite` attributes, and annotation of Id. cites falls out as well, because it's just annotating `id_cite.span()` with `my_url(id_cite.target_cite)`.

--

If it worked that way, then it might be worth tweaking `_match_shortcase_citation` and `_match_supra_citation` to scan backwards and take the _first_ of each full cite encountered, rather than removing duplicates with `list(set(full_cites))` and taking an arbitrarily chosen one. In a decision like `Majority: [full cite] [short cite] Dissent: [same full cite] [short cite]` we would expect the first short cite's target_cite to point to the first full cite, and the second to the second. Not a huge deal since it wouldn't matter for most uses.

--

You have some extra moving parts here that are serving what I think is a useful purpose I don't quite have my head around yet. :) Like there's the `Opinion/OpinionBase` abstraction that provides a caller-overridable equality between `FullCaseCitation`s, as best I can tell, but doesn't help with identity between short citations and full citations. 

I'm not super clear if that's needed, or if you could get the core of it by adding `def opinion_key(self): return (self.reporter, self.volume, self.page)` to `FullCaseCitation`, and `def opinion_key(self): return self.target_cite.opinion_key() if self.target_cite else None` to the short citation objects. You can do the backwards-scanning in `_match_shortcase_citation` by checking if `cite.opinion_key() in observed`, and the counting/clustering by making a groupby dict of `{cite.opinion_key(): [<list of cites>]}`. Does that get to the same place with less conceptual overhead, or is it not quite solving the same problem?

--

Sorry for this super late feedback, and thanks for working on this -- I'm excited to see where it's going!

mattdahl:
No worries, not late at all!

> As far as the API I feel like there might be a step before counting citations of just linking each short cite to its full cite with an attribute? Like imagine a task we wanted to do with this API was annotate each Id. cite with the URL of the full cite. Right now I think that's challenging, because you're finding but not storing the target full cite. But if the API was along the lines of ...
> 
> ```
> cites = get_citations()
> link_short_cites(cites, ...)  # Set target_cite: Optional[FullCaseCitation] attribute of each short cite, using similar logic to the current `count_citations` function
> ```
> 
> ... then counting can be done in a few lines based on `target_cite` attributes, and annotation of Id. cites falls out as well, because it's just annotating `id_cite.span()` with `my_url(id_cite.target_cite)`.

Yes, I see what you mean. My (not explained) intention in adding the `successful_match_callback` parameter was to allow for things like this. I.e., if the user wanted to do something after a match was made (such as storing the URL of the matched opinion as a new attribute), they could just pass a function that would do whatever object modification they wanted. Indeed, in CL, we would pass a function that would set (our custom) fields `match_id` and `match_url` on each citation object precisely so we could include that information in subsequent annotation, just as you propose.

> If it worked that way, then it might be worth tweaking `_match_shortcase_citation` and `_match_supra_citation` to scan backwards and take the _first_ of each full cite encountered, rather than removing duplicates with `list(set(full_cites))` and taking an arbitrarily chosen one. In a decision like `Majority: [full cite] [short cite] Dissent: [same full cite] [short cite]` we would expect the first short cite's target_cite to point to the first full cite, and the second to the second. Not a huge deal since it wouldn't matter for most uses.

That's interesting, I wasn't really thinking of it being important what the "true" target citation was. But I guess keeping that information would be helpful for adding a link that jumped to the actual up-page target? Or is there another use case?

> You have some extra moving parts here that are serving what I think is a useful purpose I don't quite have my head around yet. :) Like there's the `Opinion/OpinionBase` abstraction that provides a caller-overridable equality between `FullCaseCitation`s, as best I can tell, but doesn't help with identity between short citations and full citations.
> 
> I'm not super clear if that's needed, or if you could get the core of it by adding `def opinion_key(self): return (self.reporter, self.volume, self.page)` to `FullCaseCitation`, and `def opinion_key(self): return self.target_cite.opinion_key() if self.target_cite else None` to the short citation objects. You can do the backwards-scanning in `_match_shortcase_citation` by checking if `cite.opinion_key() in observed`, and the counting/clustering by making a groupby dict of `{cite.opinion_key(): [<list of cites>]}`. Does that get to the same place with less conceptual overhead, or is it not quite solving the same problem?

The point of the `Opinion/OpinionBase` abstraction, in my mind, was merely to provide flexibility to the user in a type-friendly way. I didn't want to define a "thick" object representation, so I created a very "thin" one that inherited from an essentially empty `OpinionBase` mixin that the user could also inherit from to make whatever opinion object they actually wanted to use satisfy the type expectations of `count_citations()`.

But, if the opinion definition is so thin, what's the point? (I think that this is a partial restatement of your question.) The alternative design that you suggest also makes sense to me, but I think there's value in keeping the idea of an actual "citation clustering" object that groups citations itself. I call it `OpinionBase/Opinion` because if it were fully built out, such an object would be an opinion, conceptually.

----

Hopefully that sheds light on my thought process, happy to continue discussing. I admit my bias in designing this was to (1) make it as easy to merge with CL as possible and (2) provide just the bare bones of matching/counting framework for the naive user. CL of course has its own matching logic (using its database), requires a certain set of information for its annotation process, uses its own opinion abstraction, etc., but I didn't want to prescriptively put any of that into eyecite, so I just tried to provide basic functionality with a lot of built-in flexibility for the sophisticated user (via the callbacks etc.).

jcushman:
Hmm, maybe clustering is the core operation here? This feels intuitive to me as a caller:

```
> cites = get_citations(...)
[full1, full2, short1, short2, full1again, short1again]

> clusters = cluster_citations(cites)
[
    [full1, short1, full1again, short1again],
    [full2, short2]
]
```

And then you can count it or build Opinions from it or generate annotations from it or whatever?

mattdahl:
Sorry for the slow reply!

I agree that it may be helpful to decouple the clustering step from the counting step. However, in that scenario, I feel like it would still be preferable to return a dictionary with "opinion clusters" as keys and lists of matched citations as values (rather than the list of lists you suggest).

It's possible we're speaking past each other because we have different ideas about what we're matching citations *to*. I think I want to match them to something like an opinion object; I think you want to match them to other citations. Does that seem right to you?

jcushman:
Hmm, yeah, that makes sense -- it took me a minute but I can see how a dictionary with opinions as keys could be handy. 

Given that I think we're headed toward extracting statutes and other things as well as opinions, maybe we get more general and just call the cited thing a "resource"?

What if I propose this terminology to explain this API:

* A "resource" is a thing that can be cited.
* "resolving" a full cite means identifying the resource it refers to.
* "resolving" a short cite means identifying the full cite it refers to.

Then the thing you're already doing can be described as: `resolve_citations` resolves each full citation into a resource object, resolves each short cite to its full cite, and returns a dictionary mapping each resolved resource to the list of citations that refer to it. I think that makes it click with my brain anyway. And returning the dictionary instead of the list I proposed earlier makes sense because then you don't throw away the work you did resolving the full cites.

---

This clicks even more for me if we don't require the caller to subclass `Opinion` (aka `Resource`) -- because the whole point of returning a dict with Opinion objects that don't do anything by default is that the caller can provide their own custom thing instead, right? So what if we just let them provide a Django model object or a tuple or whatever they want. Our API really only has to specify that a "resource" is some object that is hashable and where `A == B` if `A` and `B` are the same resource.

To play out how that would work ...

```
# complex input text with two full cites that refer to the same case, 
# an id cite, and a hypothetical non-case cite:
cites = get_citations("576 U.S. 644. 576 U.S. at 645. 135 S. Ct. 2584. Id. 17 U.S.C. § 107 (2012).")

# simple resolution without a custom resolver:
resolve_citations(cites)
{
    <resource1>: [FullCite("576 U.S. 644"), ShortCite("576 U.S. at 645")],
    <resource2>: [FullCite("135 S. Ct. 2584"), IdCite("Id.")],
    <resource3>: [StatuteCite("17 U.S.C. § 107")],  # one day ...
}

# fancy example with a custom resolver
from eyecite.cluster import default_resolve
def resolve(full_cite):
    resource = MyOpinion.objects.filter(citation__reporter=...).first()
    if not resource:
        resource = default_resolve(full_cite)
    return resource

resolve_citations(cites, resolve_full_cite=resolve)
{
    <django_opinion>: [FullCite("576 U.S. 644"), ShortCite("576 U.S. at 645"), FullCite("135 S. Ct. 2584"), IdCite("Id.")],
    <resource>: [StatuteCite("17 U.S.C. § 107")],
}
```

Behind the scenes, we just keep a `resolutions` lookup that maps full_cite -> resource, and iterate over that instead of over `opinions` when resolving short cites, and otherwise I think everything still works.

mattdahl:
Okay, yes, I think I totally agree with all of that. I will make changes accordingly and push a new commit soon. Thanks for the pseudocode, that's very helpful!

mattdahl:
Okay, I think I've made changes that reflect your described API, though I may have deviated somewhat in the implementation (e.g., I think that the `resolutions` map that I use is different than the one that you mentioned at the end). Let me know what you think!

(No tests yet until we settle on the API. Also, the linter is failing because I haven't upgraded to the new version of black that Mike was talking about in Slack. Will fix on next push...)

eyecite/models.py
```diff
@@ -403,3 +403,24 @@ def compiled_regex(self):
...
+
+
+@dataclass(frozen=True)
+class Resource:
+    """Thin resource class representing an object to which a citation can be
+    resolved."""
+
+    citation: FullCaseCitation
+
+    def __hash__(self):
```
jcushman, line 415: I think this can just be `return hash((self.citation.reporter, self.citation.volume, self.citation.page))`? There's an example at https://docs.python.org/3.5/reference/datamodel.html#object.__hash__

eyecite/resolve.py
```diff
@@ -0,0 +1,177 @@
...
+    IdCitation,
+    NonopinionCitation,
+    Resource,
+    ShortCaseCitation,
+    SupraCitation,
+)
+from eyecite.utils import strip_punct
+
+
+def _resolve_fullcase_citation(citation: FullCaseCitation) -> Resource:
```
jcushman, line 16: This might want to be named something without an underscore, since it's probably a best practice to import it and call it on unmatched citations when overriding it? Not sure about the other three, but this one feels like part of the API.

jcushman:
Cool! Other than maybe renaming`_resolve_fullcase_citation` as mentioned in my review comment, I'm happy with the public API.

Implemention-wise your internal methods like `_resolve_shortcase_citation` still assume that `_resolve_fullcase_citation` is returning a `Resource` rather than an `Any`. That's what I was getting at with the need to pass around an internal dictionary or list-of-tuples mapping full_cite -> resource -- this bit of `_resolve_shortcase_citation` needs to not assume that resources have a `citation` attr:

```
    for r in resources:
         if r.citation.defendant and ag in r.citation.defendant:
             candidates.append(r)
```

That wants to become something like

```
    for cite, resource in resolved_full_cites.items():
        if cite.defendant and ag in citation.defendant:
            candidates.append(resource)
```

One other thought about typing -- rather than typing resources as `Any`, you could do something like

```
from typing import Hashable

ResourceType = Hashable

resolve_fullcase_citation: Callable[[FullCaseCitation], ResourceType]
```

Sounds like Hashable implies `__eq__` as well as `__hash__` so that should be correct.

(I've been jamming out on this corner of Python typing this week -- python supports both class-based types like `resource: Resource` and protocol-based types like `resource: Hashable`, and the second paradigm is nice for these situations where we're embracing the duck-typing-friendly side of the language.)

mattdahl:
Ahh, I understand the intuition now. I agree that this is more elegant. I think I've got it now (see diff in 9497d05) -- the only change I believe I made was to flip the key-value pairing to (resource, citation) in the `resolved_full_cites` dict so we could continue taking advantage of the hashability of resources to automatically remove duplicates. Your suggestion about defining a custom `ResourceType` type is also great; I think that adds a lot of semantic clarity in addition to specifying the need for hashability in overrides of `Resource`.

Let me know how you think it's looking now.

jcushman:
Cool! This API is feeling right to me. I'd be fine with pulling it in at this point and iterating from there. 👍

One edge case if you key on `resource` is you can't resolve short cites that refer to a particular full cite. For example from my "fancy example with a custom resolver" code example above where parallel cites resolve to the same resource, the short cite won't match in "Foo v. Bar, 576 U.S. 644, 135 S. Ct. 2584. ... Foo, 576 U.S. at 645." Dunno if that happens much.

eyecite/resolve.py
```diff
@@ -0,0 +1,180 @@
...
+
+        # If the citation is a supra citation, try to resolve it
+        elif isinstance(citation, SupraCitation):
+            resolution = resolve_supra_citation(citation, resolved_full_cites)
+
+        # If the citation is an id citation, try to resolve it
+        elif isinstance(citation, IdCitation):
+            resolution = resolve_id_citation(citation, last_resolution)
+
+        if resolution:
```
mlissner, line 171: Won't this be undefined if the citation is a `NonOpinionCitation`?

eyecite/resolve.py
```diff
@@ -0,0 +1,180 @@
...
+
+def _resolve_id_citation(
+    citation: IdCitation,
+    last_resolution: ResourceType,
+) -> Optional[ResourceType]:
+    """
+    Resolve id citations to the resource of the previously resolved
+    citation.
+    """
+    return last_resolution
```
mlissner, line 100: Does this need a check that `last_resolution` is not `None`?

eyecite/resolve.py
mlissner: I'm not sure I follow Jack? Does this make sense to you, Matt? If so, I'm happy to just see a change or not.

eyecite/resolve.py
```diff
@@ -0,0 +1,180 @@
...
+        ag: str = strip_punct(citation.antecedent_guess)
+
+        for resource, full_citation in candidates:
+            if full_citation.defendant and ag in full_citation.defendant:
+                refined_candidates.append(resource)
+            elif full_citation.plaintiff and ag in full_citation.plaintiff:
+                refined_candidates.append(resource)
+
+        # Only accept if one candidate remains
+        return refined_candidates[0] if len(refined_candidates) == 1 else None
```
mlissner, line 60: This feels like it'd be a candidate for factoring out.

mlissner (commented):
Well, I had a busy week last week and I'm just catching up, but this early quote from me made me laugh:

>  I think it feels about right though 

Ah, past Mike, how wrong you were! Thank you both for moving this forward so much.

I've got a couple little inline comments, but this looks great aside from those. The only other general code comment I have is to wonder whether some of the resolver code would be easier to read if we used variable names like `id_cite`, `supra_cite`, and `full_cite` wherever possible?

And then we just need tests, right?

eyecite/resolve.py
jcushman: Sorry, to be less opaque, I'm referring to the PEP8 rule that variable names starting with underscore are part of the internal interface of a module and are allowed to change without warning. I don't think this method qualifies, because people should be encouraged to import it and call it from user code in the documentation. So I'd suggesting naming it something that is (a) public and (b) future-proof when we add non-case citation extraction. How about changing `_resolve_fullcase_citation` to `resolve_full_citation`?

Why public and documented? Because the reason we return a dict from `resolve_citations` is to help people who override this function, and when they do that they ought to call the base function from the override. So something like this should be encouraged in the docs:

```
from eyecite.resolve import resolve_full_citation

def my_resolve(full_cite):
    # special handling for resolution of known cases in our database
    resource = MyOpinion.objects.filter(citation__reporter=...).first()
    if resource:
        return resource
    # allow normal clustering of other citations
    return resolve_full_citation(full_cite)

resolve_citations(citations, resolve_full_citation=resolve_full_citation)
{
    <MyOpinion object>: [<full_cite>, <short_cite>, <id_cite>],
    <Resource object>: [<full cite>, <short cite>],
}
```

eyecite/resolve.py
mattdahl: I don't think so, because if `last_resolution` is `None`, then we would return `None` anyway. Thus, we can simply return `last_resolution` itself, no matter what.

eyecite/resolve.py
mattdahl: Yes, good catch.

eyecite/resolve.py
mattdahl: This seems logical to me, thanks for explaining more.

eyecite/resolve.py
mattdahl: I see what you mean, but can you propose a name for the sub-function? I'm not quite sure what to call it.

mattdahl:
Thanks for the comments, all. 9fd1d5a should incorporate all your suggestions, including Jack's point that keying on resources in the interal `resolved_full_cites` dict would cause. parallel citation resolution to fail. Let me know if I missed anything.

In 8eb7d46 I have added some tests, though I am open to suggestions about how to make them more DRY. The problem is that we need to repeat the same mock citation in both the `Resource` key and in the values list. I couldn't come up with an elegant way to do that without just literally repeating the mock citation call, which seems repetitive.

Also, tests are failing on GitHub but I'm not sure why. Something is wrong with `flake8`?

eyecite/models.py
jcushman: Did this not pan out?

jcushman:
Since this has been open for a long time and the API seems settled, I'd be happy with pulling it in and iterating as needed from there. 👍

> In 8eb7d46 I have added some tests, though I am open to suggestions about how to make them more DRY.

Maybe just go back to regular test methods for this?

```
def test_multiple_full_cites(self):
    cite1 = case_citation(0)
    cite2 = case_citation(4, reporter="F.2d")
    self.assertResolution(
        [cite1, cite2], 
        {Resource(cite1): [cite1], Resource(cite2): [cite2]}
    )

def test_supra_cite(self):
    cite = case_citation(0, defendant="Bar")
    supra = supra_citation(0, antecedent_guess="Bar")
    self.assertResolution(
        [cite, supra], 
        {Resource(cite): [cite, supra]}
    )
```

And put the inside of your test loop in an `assertResolution` helper method.

eyecite/resolve.py
mlissner: Understood the underscore part, but didn't think calling the function at the end of the override would be useful or standard, but happy to be wrong. Resolving.

eyecite/resolve.py
mlissner: Something like, `filter_by_matching_antecedent`?

mlissner:
Just the one optional tweak from me re factoring out a function, and fixing tests, but everything else looks good.

eyecite/models.py
mattdahl: My bad, I thought I made this change but I guess I didn't. Forthcoming.

mattdahl:
All right, I refactored that bit that Mike pointed out, made the hash function change suggested earlier by Jack, and reformatted all the tests to be somewhat less verbose. Also updated the README accordingly.

I rebased for the sake of keeping the commit history on master clean. I think this should be ready to merge (I don't think the `flake8` issue with the GitHub checks is my fault), but let me know if anything seems outstanding.

mlissner:
Merged! Thank you both for all your work on this. It's a great new feature. Matt, the failing tests were because the CI caches were failing. I still haven't figured out what that happens sometimes, but I just changed their keys and that fixed it. 

Anyway, great stuff. Thank you!

jcushman:
This is great! Thanks for plowing through all my ~bikeshedding~ constructive feedback, Matt!

mattdahl:
Ha, hardly bikeshedding. The code is much improved because of your comments.

PR #59, opened 2021-04-19
Update year parsing

jcushman:
This updates year parsing to fix a ValueError CAP ran into on the input "123 U.S. 456 (196⁴)", and makes some other changes along the way.

* No longer use `isdigit`, which I just learned returns True for unicode digit characters, just attempt int() conversion and catch ValueError.
* Don't accept all strings that start with four digit characters, so we don't parse `"2000000"` as `2000`. If there are things we're trying to handle with `r"(\d{4}).*"` let's use a tighter regex.
* Bump the minimum date back to 1600, since CAP goes back to 1658.
* Add a maximum date of the current year.

### DIFF
--- a/eyecite/helpers.py
+++ b/eyecite/helpers.py
@@ -1,4 +1,5 @@
 import re
+from datetime import date
 from typing import List, Optional, Union
 
 from courts_db import courts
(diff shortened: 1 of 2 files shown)


eyecite/helpers.py
```diff
@@ -48,20 +49,17 @@ def get_court_by_paren(
     return court_code
 
 
+_this_year = date.today().year
```
mlissner, line 52: I'm pretty sure we get cases from next year sometimes from weird courts in late December. 

Maybe:

```suggestion
_next_year = date.today().year + 1
```

jcushman:
Good catch, I also added a comment.

Issue #56, opened 2021-04-14
Tighten matches for roman numeral page numbers


jcushman:
Once I get eyecite running on the CAP corpus, I'd like to look at all of the extracted citations where the page number was supposedly a roman numeral, and figure out what the error rate is and if we can filter out common false matches or restrict roman numeral matches to some reporters and volumes.

Here are false positive matches I've collected so far:

- **49 Col. L.** Rev. 875
- **65 Mich. L.** Rev. 477
- **37 Taylor v.** Bd. of Educ., 240 Fed. Appx. 717 (6th Cir. 2007)
- **11 U. S. C.** § 701

(Moving this conversation over from #18 to focus specifically on roman numerals.)

jcushman:
OK, I now have the data for running eyecite on CAP. I think the upshot is we (CAP at least) don't want to look for roman numeral page numbers at all, unless they match a list of known volumes with cases indexed by roman numeral. Here's the details ...

CAP has roughly 100 million full or short cites, of which roughly 500,000 have roman numeral page numbers. About 95% of those seem to be false positives.

Here are the ten most common roman numeral cites:

total | 501671
-- | --
28 U. S. C | 13879
42 U. S. C | 9941
18 U. S. C | 7975
42 Pa. C | 7191
29 U. S. C | 6345
28 U.S. C | 5668
15 U. S. C | 5333
6 R. C | 4881
26 U. S. C | 4009

All data: [roman_numerals_counts.csv](https://github.com/freelawproject/eyecite/files/6327154/roman_numerals_counts.csv)

Those are all false positives for "C", but looking down the list there's a lot of other common false positives as well -- "10 A. M", "7 H. L", "32 Hen. VIII" etc. (That last one is awesome! We're finding cites to the law of King Henry VIII.)

Here are counts for the top page numbers:

total | 501656
-- | --
C | 279821
L | 104392
l | 35970
M | 34795
D | 9512
v | 5608
I | 3740
m | 1180
VIII | 1163
x | 1077
vii | 957
xi | 911

All data: [roman_numerals_pages.csv](https://github.com/freelawproject/eyecite/files/6327166/roman_numerals_pages.csv)

I think these are basically all false positives down to `x` or `vii` or so. If we assume that all cites are false positives where `page in ['C', 'L', 'M', 'D', 'v', 'l', 'm', 'did']` or `reporter in ['Edw.', 'Hen.', 'Rich.', 'Will.', 'H.']` (the English kings), we end up with this:

[...]

Issue #52, opened 2021-03-31
Correct way to handle text abutting citation?


jcushman:
What do we think is the right thing to do with text like this?

```
In [2]: get_citations("foo1 U.S. 1bar")
Out[2]: [FullCaseCitation(token=CitationToken(data='1 U.S. 1bar', start=3, end=14, volume='1', reporter='U.S.', page='1bar'...
```

We're currently permissive about finding cites with abutting characters, which I can imagine being overinclusive with reporter strings like "foo1 Or 2bar" or "foo1 P 2bar" or "foo1 mt 2bar" etc.

I guess this breaks down into a few separate things:

1. Should there be a defined list of characters that can come before the start of a cite, like `(^| )`? I'm thinking probably yes?
1. Should there be a defined list of characters that can come after a cite? This seems harder to make complete, but might be like `($|[ ,.;])` and a few more along those lines. Less sure about that one.
1. Should we narrow the definition of page numbers so that example would come back as `page='1'`? We currently capture `1bar` because of the page number regex `\d{1,6}[-]?[a-zA-Z]{1,6}`, which is there to handle `A special Connecticut or Illinois number. E.g., "13301-M"`. So we could add that regex specifically to CT and IL in reporters-db and avoid capturing random cruft in other reporters. I'm leaning toward that being a good idea.

mlissner:
I guess our bias should be towards underinclusive correctness. If we start from there, I think the best options are numbers 1 and 3. Option 1, tightening what can come before a citation, seems easy. Option 2, tightening what comes after, seems hard. Option 3, tightening our current regex and then dialing in special cases, seems not too bad either.

mlissner:
This is fixed via #61, right @jcushman ?

jcushman:
Yep!

PR #61, opened 2021-04-26
Tighten page number matching; extend pin cite matching; extract parentheticals

jcushman:
As usual I started pulling a string and ended up reknitting the sweater, so this is way too big to review -- apologies yet again. :)

This PR grows out of #56, regarding excess roman numeral page matches, and #52, regarding tightening full-cite matches to not match patterns like "A1 P 1A" as "1 P. 1". Along the way to getting those working I ended up building a useful helper function for running regular expressions against the extracted tokens, and using it to more precisely extract pin cites, and I threw in a bonus feature to extract citation parentheticals as well.

**TL;DR -- API changes:**

Here's what I know of that visibly changes with this PR when calling `get_citations()`:

* Attribute changes:
  * All citations gain a `pin_cite` attribute, which includes the entire pin cite ("at 1, 2-3 & n. 7").
  * `IdCitation` loses the `has_page` and `after_tokens` attributes.
  * `SupraCitation` loses the `page` attribute.
  * (`ShortCaseCitation` keeps its `page` attribute, so "1 U.S. at 4-5" ends up with `page="4", pin_cite="4-5"`. I don't know if this is correct or if page should be `None`, but it feels reasonable since we at least know that short cites ought to start with a page number.)
  * `FullCaseCitation` gains a `parenthetical` attribute.
  * `antecedent` no longer includes the trailing comma.
  * The highlighted `span()` range for `IdCitation`, `SupraCitation`, and `ShortCaseCitation` now includes the pin cite.
* Matching restrictions:
  * Cites no longer match with trailing or following alphanumeric characters, like "A1 P 1" or "1 P 1A".
  * Metadata extraction joins cite extraction in no longer extending across paragraph breaks, so "Id.\n2" won't be treated as "Id. at 2".

**Narrative, for code review:**

Here's how all of this ended up in one bundle ...

* First let's distinguish between page numbers and pin cites: a page number is part of the official citation of a case, like "1 U.S. 1". A pin cite is a string of one or multiple tokens giving extra metadata about where within a document the citation refers to, like "1 U.S. 1, **4-5**", or "1 U.S., **at 7-8 & n. 3**", or "Foo, supra, **at ¶10-11**", or "Id. **at 1:10-2:5**". This whole massive PR relates to detangling those two concepts so we can more precisely identify actual case cites. :)
  * Query: does "at " belong in the `pin_cite` attribute for those examples? I could go either way. It's currently in.
* So first we want to tighten the `PAGE_NUMBER_REGEX` for matching actual page numbers. Almost all of that existing regex  has moved to reporter-specific regexes in reporters-db and can be deleted. The one exception is roman numerals, which we do want to match since there are frequent-enough citations to front matter. But per #56 we only want to match lower case numerals below 200 other than `v`, `l`, and `c` to avoid false positives, so let's add a fancy regex to do that and a test to make sure it's correct.
* Per #52, we'll also tighten the full cite token extractor to require non-alphanumeric characters to the left and right of the citation itself.
  * This means we no longer need `remove_address_citations`, so remove that.
* Now that we've tightened cite extraction this way, our tests expecting "1 U. S., at 20-25" to come in with the page "20-25" start to fail! That test _should_ fail -- it's not a page, it's a pin cite for a case named "1 U.S. 15" or something. So we add a big `PIN_CITE_REGEX` that knows how to match all of those pin cite formats. And we need some way to run it, so we'll add a function `match_on_tokens` that is able to scan through `words = [ShortCite("1 U.S., at 20"), "-", "25"]` and build up a string like "20-25" to apply the pin-cite-matching regex to. Once we're able to extract data with regexes forwards and backwards from a given token, we can use that to attach a new `pin_cite` metadata attribute to short cites and supra cites and id. cites.
  * This new regex scanning approach with `match_on_tokens` allows our pin cite matching to be much more robust! Where before we were matching just "Id. at <digit>" or "Id. <digit>", we'll now get stuff like "Id. at 3, 4-5 & n. 7".
* Uh oh, another test failed -- when we later insert annotations we want to link to "Id. at 2" instead of just "Id.", so cite objects need to keep track of where they came from. `IdCitation` was doing that as a special case with the `has_page` and `after_tokens` attributes, but we're not inspecting tokens that way anymore, so let's add `span_start` and `span_end` attributes to all citations to let us override the highlighted text if need be, and use that to extend the highlight to include the pin cite for `IdCitation`, `SupraCitation`, and ShortCitation`.
  * For character-level positioning like this to work, we have to start keeping track of spaces between words. Let's update `Tokenizer` to use `split(" ")` and include the spaces in the emitted tokens with a new `append_text()` method, rather than using `split()` and throwing out all whitespace.
  * For _this_ to work, we have to start tracking paragraph breaks separately, so add a `ParagraphToken` to extract newlines. This turns out to be an accuracy win, because it means we won't interpret "Id.\n2" as "Id. at 2" like we used to. (Remember this is consistent with the existing rule that callers are required to replace newlines with spaces if they don't know where the real paragraph breaks are in their document.)
* Now that we have `match_on_tokens` working, we can use it in some other places!
  * We can do a better job of extracting pin cites and other metadata for full citations like "1 U.S. 1, 4-5, 2 S. Ct. 2, 6-7 (4th Cir. 2012) (overruling foo)". Let's use `match_on_tokens` in `add_post_citation()` to handle that. Now if we're annotating "1 U.S. 1", we'll get "4-5" as `pin_cite` rather than as part of `extra`, and with one more line of the regex we'll pull in "overruling foo" as the `parenthetical` as well.
  * Let's use `match_on_tokens` to extract the antecedent case name for `ShortCite` and `SupraCite`. This currently works the same as it did before, but it will make it easier to use a more flexible pattern in the future.
  * This means the one thing we're _not_ currently using a regex for is `add_defendant()` -- that one's still scanning token by token until it hits a stop word. I'm not sure if that wants to stay that way, but I left it for now because I don't have a strong idea of how to improve case name detection yet, and it's fine to use both strategies for now.
* Getting the new whitespace handling to work took some profiling, in the process of which I noticed that `get_citations()` was spending a lot of time skipping over string tokens. For efficiency let's update `tokenize()` to return separate lists of all tokens and just non-string tokens.

### DIFF
--- a/eyecite/find_citations.py
+++ b/eyecite/find_citations.py
@@ -1,12 +1,14 @@
-from typing import Iterable, List, Optional, cast
+from typing import Iterable, List, cast
 
 from eyecite.helpers import (
+    SHORT_CITE_ANTECEDENT_REGEX,
+    SUPRA_ANTECEDENT_REGEX,
     add_defendant,
     add_post_citation,
     disambiguate_reporters,
+    extract_pin_cite,
     joke_cite,
-    parse_page,
-    remove_address_citations,
+    match_on_tokens,
 )
 from eyecite.models import (
     CitationBase,
@@ -19,7 +21,6 @@ from eyecite.models import (
     ShortCaseCitation,
     SupraCitation,
     SupraToken,
-    Token,
     Tokens,
 )
 from eyecite.tokenizers import Tokenizer, default_tokenizer
@@ -36,11 +37,11 @@ def get_citations(
     if plain_text == "eyecite":
         return joke_cite
 
-    words = cast(Tokens, list(tokenizer.tokenize(plain_text)))
+    words, citation_tokens = tokenizer.tokenize(plain_text)
     citations: List[CitationBase] = []
 
-    for i, token in enumerate(words):
-        citation: Optional[CitationBase]
+    for i, token in citation_tokens:
+        citation: CitationBase
         token_type = type(token)
 
         # CASE 1: Citation token is a reporter (e.g., "U. S.").
@@ -71,7 +72,7 @@ def get_citations(
         # It could be any of the previous citations above. Thus, like an Id.
         # citation, for safety we won't resolve this reference yet.
         elif token_type is SupraToken:
-            citation = extract_supra_citation(cast(SupraToken, words), i)
+            citation = extract_supra_citation(words, i)
 
         # CASE 4: Citation token is a section marker.
         # In this case, it's likely that this is a reference to a non-
@@ -84,15 +85,12 @@ def get_citations(
         else:
             continue
 
-        if citation is not None:
-            citations.append(citation)
+        citations.append(citation)
 
     # Remove citations with multiple reporter candidates where we couldn't
     # guess correct reporter
     if remove_ambiguous:
         citations = disambiguate_reporters(citations)
-
-    citations = remove_address_citations(citations)
 
     # Returns a list of citations ordered in the sequence that they appear in
     # the document. The ordering of this list is important for reconstructing
@@ -132,19 +130,22 @@ def extract_shortform_citation(
     Shortform 1: Adarand, 515 U.S., at 241
     Shortform 2: 515 U.S., at 241
     """
-    # Get antecedent -- either previous word, or previous two words if
-    # previous word is a comma
+    # get antecedent word
     antecedent_guess = None
-    if index > 0:
-        antecedent_guess = str(words[index - 1])
-        if antecedent_guess == ",":
-            if index > 1:
-                antecedent_guess = str(words[index - 2]) + ","
-            else:
-                antecedent_guess = None
+    m = match_on_tokens(
+        words,
+        index - 1,
+        SHORT_CITE_ANTECEDENT_REGEX,
+        strings_only=True,
+        forward=False,
+    )
+    if m:
+        antecedent_guess = m[1].strip()
 
     # Get citation
     cite_token = cast(CitationToken, words[index])
+
+    pin_cite, span_end = extract_pin_cite(words, index, prefix=cite_token.page)
 
     # Return ShortCaseCitation
     return ShortCaseCitation(
@@ -157,13 +158,15 @@ def extract_shortform_citation(
         reporter_found=cite_token.reporter,
         exact_editions=cite_token.exact_editions,
         variation_editions=cite_token.variation_editions,
+        pin_cite=pin_cite,
+        span_end=span_end,
     )
 
 
 def extract_supra_citation(
     words: Tokens,
     index: int,
-) -> Optional[SupraCitation]:
+) -> SupraCitation:
     """Given a list of words and the index of a supra token, look before
     and after to see if this is a supra citation. If found, construct
     and return a SupraCitation object.
@@ -173,37 +176,32 @@ def extract_supra_citation(
     Supra 3: Adarand, supra, somethingelse
     Supra 4: Adrand, supra. somethingelse
     """
-    # Get page, with bounds check to ensure we don't scan past end of words.
-    # Use index + 2 to skip "at".
-    page = None
-    if index + 2 < len(words):
-        page = parse_page(str(words[index + 2]))
-
-    # Get antecedent -- either previous word, or previous two words if
-    # previous word is a comma. If previous word is a digit, store as
-    # volume and use next previous word as antecedent.
+    pin_cite, span_end = extract_pin_cite(words, index)
     antecedent_guess = None
     volume = None
-    if index > 0:
-        antecedent_guess = str(words[index - 1])
-        if antecedent_guess.isdigit():
-            volume = antecedent_guess
-            if index > 1:
-                antecedent_guess = str(words[index - 2])
-            else:
-                antecedent_guess = None
-        elif antecedent_guess == ",":
-            if index > 1:
-                antecedent_guess = str(words[index - 2]) + ","
-            else:
-                antecedent_guess = None
+    m = match_on_tokens(
+        words,
+        index - 1,
+        SUPRA_ANTECEDENT_REGEX,
+        strings_only=True,
+        forward=False,
+    )
+    if m:
+        if m[1]:
+            antecedent_guess = m[1].strip()
+            volume = m[2].strip()
+        elif m[3]:
+            volume = m[3].strip()
+        else:
+            antecedent_guess = m[4].strip()
 
     # Return SupraCitation
     return SupraCitation(
         cast(SupraToken, words[index]),
         index,
-        antecedent_guess,
-        page=page,
+        span_end=span_end,
+        antecedent_guess=antecedent_guess,
+        pin_cite=pin_cite,
         volume=volume,
     )
 
@@ -211,48 +209,15 @@ def extract_supra_citation(
 def extract_id_citation(
     words: Tokens,
     index: int,
-) -> Optional[IdCitation]:
+) -> IdCitation:
     """Given a list of words and the index of an id token, gather the
     immediately succeeding tokens to construct and return an IdCitation
     object.
     """
-    # List of literals that could come after an id token
-    id_reference_token_literals = {
-        "at",
-        "p.",
-        "p",
-        "pp.",
-        "p",
-        "@",
-        "pg",
-        "pg.",
-        "¶",
-        "¶¶",
-    }
-
-    # Helper function to see whether a token qualifies as a page candidate
-    def is_page_candidate(token):
-        return token in id_reference_token_literals or (
-            not isinstance(token, Token) and parse_page(token)
-        )
-
-    # Check if the post-id token is indeed a page candidate.
-    # If so, set scan_index to capture all page candidates immediately
-    # following Id. cite.
-    scan_index = index + 1
-    has_page = False
-    while scan_index < len(words) and is_page_candidate(words[scan_index]):
-        scan_index += 1
-        has_page = True
-
-    # If it is not, simply set a naive anchor for the end of the scan_index.
-    if not has_page:
-        scan_index = min(index + 4, len(words))
-
-    # Only linkify the after tokens if a page is found
+    pin_cite, span_end = extract_pin_cite(words, index)
     return IdCitation(
         cast(IdToken, words[index]),
         index,
-        after_tokens=words[index + 1 : scan_index],
-        has_page=has_page,
+        span_end=span_end,
+        pin_cite=pin_cite,
     )
--- a/eyecite/find_citations.py
+++ b/eyecite/find_citations.py
@@ -140,7 +140,7 @@ def extract_shortform_citation(
         forward=False,
     )
     if m:
-        antecedent_guess = m[1].strip()
+        antecedent_guess = m["antecedent"].strip()
 
     # Get citation
     cite_token = cast(CitationToken, words[index])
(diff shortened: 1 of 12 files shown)


jcushman:
@mattdahl I've been meaning to take a look at your cite-attribution PR but haven't yet! This might relate at least inasmuch as messing a little with the API for getting page numbers from short cites.

tests/test_UtilsTest.py
```diff
@@ -30,3 +33,18 @@ def test_clean_text(self):
...
+        except 5, 50, 100."""
+        expected = (
+            list(range(1, 5))
+            + list(range(6, 50))
+            + list(range(51, 100))
+            + list(range(101, 200))
+        )
+        actual = sorted(
+            roman.fromRoman(n.upper())
+            for n in exrex.generate(ROMAN_NUMERAL_REGEX)
```
mlissner, line 48: Looks like generate is limited to 20 by default. Do we want to set the limit high enough that it'd generate all the needed regexes?

jcushman: That bit me too, but this test passes because `limit` refers to length of wildcards, not number of examples generated. `exrex.generate('[abc]+', limit=5)` will exhaustively generate all strings of `[abc]` up to length 5.

eyecite/find_citations.py
```diff
@@ -173,86 +176,48 @@ def extract_supra_citation(
...
-            else:
-                antecedent_guess = None
+    m = match_on_tokens(
+        words,
+        index - 1,
+        SUPRA_ANTECEDENT_REGEX,
+        strings_only=True,
+        forward=False,
+    )
+    if m:
```
mattdahl, line 189: This is a little opaque, maybe add a short comment explaining that here we're checking if the antecedent could be a volume number instead of just a word? (This is clear in the regex definition in`helpers.py`, but not here.)

eyecite/helpers.py
```diff
@@ -1,21 +1,115 @@
...
+#   before less specific.
+
+# Pin cite regex:
+# A pin cite is the part of a citation used to specify a particular section of
+# the referenced document. These may have prefixes, may include paragraph,
+# page, or line references, and may have multiple ranges specified.
+# For some examples see
+# https://github.com/freelawproject/courtlistener/issues/1344#issuecomment-662994948
+PIN_CITE_TOKEN_REGEX = r"""
+    # optional label -- ¶, §, p., pp., pg., n., note, & n.
```
mattdahl, line 32: Maybe add `fn` here (sometimes footnote abbreviation)?

eyecite/helpers.py
```diff
@@ -1,21 +1,115 @@
...
+# Capture metadata after a full cite. For example given the citation "1 U.S. 1"
+# with the following text:
+#   1 U.S. 1, 4-5, 2 S. Ct. 2, 6-7 (4th Cir. 2012) (overruling foo)
+# we want to capture:
+#   pin_cite = 4-5
+#   extra = 2 S. Ct. 2, 6-7
+#   court = 4th Cir.
+#   year = 2012
+#   parenthetical = overruling foo
+POST_CITATION_REGEX = rf"""
```
mattdahl, line 96: Would it be helpful to add names to these match groups? Not sure if that would make the extraction clearer below (lines 170ff.), instead of doing a bunch of `m[]` calls with the group indices?

eyecite/models.py
```diff
@@ -50,14 +50,22 @@ class CitationBase:
...
 
     def span(self):
         """Start and stop offsets in source text for matched_text.()"""
-        return (self.token.start, self.token.end)
+        return (
+            self.span_start
+            if self.span_start is not None
+            else self.token.start,
+            self.span_end if self.span_end is not None else self.token.end,
+        )
```
mattdahl, line 68: More elegant to say `return (self.span_start or self.token.start, self.span_end or self.token.end)`?

tests/test_FindTest.py
```diff
@@ -176,8 +186,8 @@ def test_find_citations(self):
             ('2006-Ohio-2095',
              [case_citation(0, source_text='2006-Ohio-2095', page='2095',
                             reporter='Ohio', volume='2006')]),
-            ('2017 IL App (4th) 160407WC',
-             [case_citation(0, page='160407WC', reporter='IL App (4th)',
+            ('2017 IL App (4th) 160407',
+             [case_citation(0, page='160407', reporter='IL App (4th)',
```
mattdahl, line 190: So I get that this is changed because we no longer allow text that abuts a citation, but is `160407WC` actually a valid page in `IL App`? I have no idea, but if so, is that something that we would put in `reporters_db` now? (I haven't been following too closely what's been going on over there.)

tests/test_FindTest.py
```diff
@@ -190,122 +200,112 @@ def test_find_citations(self):
             ('before asdf, 1 U. S., at 2',
              [case_citation(2, page='2', reporter='U.S.',
                             reporter_found='U. S.', short=True,
-                            antecedent_guess='asdf,')]),
+                            antecedent_guess='asdf')]),
             # Test short form citation with preceding ASCII quotation
             ('before asdf,” 1 U. S., at 2',
              [case_citation(2, page='2', reporter_found='U. S.',
-                            short=True, antecedent_guess='asdf,”')]),
+                            short=True)]),
```
mattdahl, line 207: Why is there no longer an antecedent detected for this one?

tests/test_FindTest.py
```diff
@@ -314,12 +314,10 @@ def test_find_citations(self):
...
              [],),
-            # Test that the tokenizer handles whitespace well. In the past, the
-            # capital letter P in 5243-P matched the abbreviation for the Pacific
-            # reporter ("P"), and the tokenizing would be wrong.
-            ('Failed to recognize 1993 Ct. Sup. 5243-P',
+            # Test Conn. Super. Ct. regex variation.
+            ('Failed to recognize 1993 Conn. Super. Ct. 5243-P',
              [case_citation(3, volume='1993', reporter='Conn. Super. Ct.',
-                            reporter_found='Ct. Sup.', page='5243-P')]),
+                            page='5243-P')]),
```
mattdahl, line 320: Would `1993 Ct. Sup. 5243-P` no longer be extracted? (Is `5243-P` only a valid page number for `Conn. Super. Ct.` and not `Ct. Sup.`?

mattdahl:
Hey, sorry for the delay if you were awaiting my comments. It looks pretty sweet, I have nothing major to suggest -- just the questions noted in-line above.

jcushman:
Thanks for the review, Matt!

There's a couple of deserved comments in there about how it's hard to read the `if m[0]` stuff. Question for @mlissner as well as @mattdahl -- how would we feel about adding a dependency on the `regex` library, which is a fancier drop-in replacement for the builtin `re` library? The thing that's making this code hard to read is that the builtin `re` doesn't let you use the same named capture group more than once, while `regex` does what you would hope and just returns the one that actually matched:

```
# with re:
import re
SUPRA_ANTECEDENT_REGEX = r"""
    (?:
        ([\w\-.]+),?\ (\d+)|
        (\d+)|
        ([\w\-.]+),?
    )
    \   # final space
"""
m = re.search(SUPRA_ANTECEDENT_REGEX, ...)
if m:
    if m[1]:
        antecedent_guess = m[1].strip()
        volume = m[2].strip()
    elif m[3]:
        volume = m[3].strip()
    else:
        antecedent_guess = m[4].strip()

# with regex:
import regex as re
SUPRA_ANTECEDENT_REGEX = r"""
    (?:
        (?P<antecedent>[\w\-.]+),?\ (?P<volume>\d+)|
        (?P<volume>\d+)|
        (?P<antecedent>[\w\-.]+),?
    )
    \   # final space
"""
m = re.search(SUPRA_ANTECEDENT_REGEX, ...)
if m:
    antecedent_guess = m['antecedent']
    volume = m['volume']
```

I have no opinion here about whether `regex` is a reasonable dependency to add, but that sure is nicer.

If we don't like the dependency, we could maybe get to the same place with helper functions that inserted suffixes to the named capture groups and merged them back together, though it adds some conceptual overhead:

```
m = re.search(expand_capture_groups(SUPRA_ANTECEDENT_REGEX), ...)
if m:
    groups = merge_capture_groups(m)
    antecedent_guess = groups['antecedent']
    volume = groups['volume']
```

... where `expand_capture_groups` replaces `<volume>` with `<volume1>, <volume2> ...` and `merge_capture_groups` converts `groupdict()` from `{'volume1': None, 'volume2': 'foo'}` to `{'volume': 'foo'}`.

eyecite/models.py
jcushman: That won't work for `self.span_start == 0` -- might work for `self.span_end` but probably clearer to keep it consistent.

tests/test_FindTest.py
jcushman: Yeah, this is handled in reporters-db [now](https://github.com/freelawproject/reporters-db/pull/47) and should work -- it's just not on PyPI yet.

tests/test_FindTest.py
jcushman: The antecedent word now has to match `([\w\-.]+),?`.

Possibly we'll find other things that have to be added to that, but I think it's correct not to include trailing quotes -- `asdf` isn't the short name of 1 U.S. in that test, is it?

tests/test_FindTest.py
jcushman: Like `160407WC`, `1993 Ct. Sup. 5243-P` should work with the version of reporters-db on master -- [here's the PR](https://github.com/freelawproject/reporters-db/pull/46).

mlissner:
I was happy to do include that regex library until I saw that its readme on PyPi was a mile long. Can't put my finger on it, but it feels like too big of a dependency for something that mostly works already. I'd go for it if somebody told me the rest of the Python world had already moved to it though. (I don't deny it enhances the code.)

Only a soft feeling. If you want it, let's do it.

mlissner:
I also just noticed no changelog: https://bitbucket.org/mrabarnett/mrab-regex/issues/406/request-for-changelog-or-similar

jcushman:
Huh, I also had no sense of whether this was a random project or what, but apparently Guido tried to get `regex` added to the standard library to replace `re` (I found messages about it as far back as 2011), and the author eventually declined in 2017 because it was easier to keep it separate:

https://mail.python.org/pipermail/python-dev/2017-October/150020.html

tests/test_FindTest.py
mattdahl: Ah yes, I see, that makes sense!

eyecite/models.py
mattdahl: Good point.

mattdahl:
It seems pretty popular according to this: https://pypistats.org/packages/regex. I'm not really dependency-averse, so I would lean towards using it, but that's just my two cents.

eyecite/find_citations.py
jcushman: Fixed with regex library.

jcushman:
OK, I added the `regex` dependency -- if features haven't been added to the `re` library for the last 10 years because the core team figures we'll just use `regex`, then I guess we should. Turns out `regex` was already a dependency of `black`, so this makes it a main dependency instead of a dev dependency.

I think this is ready to go?

mlissner:
Looks good to me. I haven't carefully reviewed the latest, but I think we're fully baked.

mlissner:
Thank you both again!

PR #62, opened 2021-05-12
Improve parenthetical functionality and resolve silent test failures

### LINKED ISSUE #66: Trailing comma in defendant field of full case citation
While writing some tests to expose the issues pointed out by @jcushman in #62, I noticed that eyecite was capturing the comma separating the defendant of a case and the reporter citation and making it part of the defendant string. Checked some real text we're working with, same issue.

Here's a minimum reproducible example:
```python
import eyecite

input_str = 'foo v. bar, 1 U.S. 1 (2021)'
citations = eyecite.get_citations(input_str)
print(citations[0].defendant) # Prints 'bar,'
```

If this behavior isn't intended and we can safely assume that party names don't end in commas, it can be relatively trivially fixed by simply stripping commas (as well as whitespace) off the defendant name:

https://github.com/freelawproject/eyecite/blob/586cbb481e17a5e02f84a280601bd0e60acfa6c4/eyecite/helpers.py#L121-L123
```python
from string import whitespace
# ...
citation.metadata.defendant = "".join(
    str(w) for w in words[start_index : citation.index]
).strip(whitespace + ',')
```

Happy to create a new PR with this fix and a test to cover it, or fold it into #62. Just wanted to check if this is on purpose first.

### LINKED ISSUE #67: Silent FindTest test failures
Sorry for the spam today, but this one seems pretty important. While rebasing my code on master to incorporate the #64 refactor, I noticed my tests that should be failing were suddenly passing.

The problem (introduced in #64), it would seem, is forgetting to return a value from `get_comparison_attrs`, resulting in the test runner asserting `None == None` repeatedly and passing. To illustrate this, here's a commit that adds an obviously wrong test that should fail instead passing all CI checks: https://github.com/legal-network/eyecite/commit/bcf636799c9d51e70e291303e5a0d6c595af98b9

And here's a commit that comments out that intentionally broken test and fixes `get_comparison_attrs` so that assertions are actually made: https://github.com/legal-network/eyecite/commit/f0bc7fe2f5ac22a55226f35d768d56cee0a9e162

Fixing the method results in a significant number of test failures, most of which seem to be related to an `index` mismatch.

ProbablyFaiz:
Note: This description has been rewritten to reflect the new contents of this PR.

This PR improves and expands parenthetical support. It also partially fixes an issue in the test code causing tests to silently fail. Specifically, it does the following things:
- Modifies the parenthetical matching regex to allow for proper capturing of parentheticals with nested parentheticals.
  - For example: a citation like 'foo v. bar, 1 U.S. 3 (2021) (quoting baz (Holmes, J., concurring) and) 
  - Because we no longer cut off matching at the first closing paren found in the string, the regex could potentially capture more than one parenthetical (e.g. the parenthetical for `foo v. bar, 1 U.S. 3 (2021) (overruling baz) and something else (unrelated parenthetical).` would be captured as `overruling baz) and something else (unrelated parenthetical`.) This is addressed by adding the `process_parenthetical` helper method which counts parentheses to ensure only the first parenthetical is captured.
- Adds support for grabbing parentheticals on short form citations.\* Implements a new regex to match both the pin citation and parenthetical of these citations.
- Partially fixes the silent test failures described in #67. The silent failure is fixed by returning a value from `get_comparison_attrs`. Previously-silent failures related to a missing `court` metadata value are fixed in https://github.com/freelawproject/eyecite/pull/62/commits/a50c8ebb8015421aa2b5102ca5acbafa45c6955e, but because there are a substantial number of index mismatches and I don't currently understand the index logic, I disabled index comparison with a comment so that it can be fixed later. In the meantime, we at least get useful CI for everything else. Closes #67.
- Strips trailing commas off of defendant values. Closes #66.
- Adds a ton of tests to ensure this functionality works correctly and does not overzealously or underzealously capture parentheticals.

This PR implements a few changes that are only loosely related, so I'm happy to split this out into multiple PRs if necessary.

\* One upshot of the way I implemented this functionality in `extract_pin_citation` is that parentheticals attached to Id and Supra citations are now also captured. They are not used, however, because the Id and Supra citation classes don't currently have a parenthetical field in their metadata. I'm a little wary of tampering with a class model I don't fully understand at the moment, so I'll leave that to you guys.

### DIFF
--- a/tests/test_FindTest.py
+++ b/tests/test_FindTest.py
@@ -38,13 +38,16 @@ class FindTest(TestCase):
     def run_test_pairs(self, test_pairs, message, tokenizers=None):
         def get_comparison_attrs(cite):
             out = {
-                "index": cite.index,
+                # Indexes are all incorrect/broken at the moment, commenting this
+                # until fixed so that other functionality can be verified by CI.
+                # "index": cite.index,
                 "groups": cite.groups,
                 "metadata": cite.metadata,
             }
             if isinstance(cite, ResourceCitation):
                 out["year"] = cite.year
                 out["corrected_reporter"] = cite.corrected_reporter()
+            return out
 
         if tokenizers is None:
             tokenizers = tested_tokenizers
--- a/tests/test_FindTest.py
+++ b/tests/test_FindTest.py
@@ -201,25 +201,30 @@ class FindTest(TestCase):
             # Test first kind of short form citation (meaningless antecedent)
             ('before asdf 1 U. S., at 2',
              [case_citation(2, page='2', reporter_found='U. S.', short=True,
-                            metadata={'antecedent_guess': 'asdf'})]),
+                            metadata={'antecedent_guess': 'asdf',
+                                      'court': 'scotus'})]),
             # Test second kind of short form citation (meaningful antecedent)
             ('before asdf, 1 U. S., at 2',
              [case_citation(2, page='2', reporter='U.S.',
                             reporter_found='U. S.', short=True,
-                            metadata={'antecedent_guess': 'asdf'})]),
+                            metadata={'antecedent_guess': 'asdf',
+                                      'court': 'scotus'})]),
             # Test short form citation with preceding ASCII quotation
             ('before asdf,” 1 U. S., at 2',
              [case_citation(2, page='2', reporter_found='U. S.',
-                            short=True)]),
+                            short=True,
+                            metadata={'court': 'scotus'})]),
             # Test short form citation when case name looks like a reporter
             ('before Johnson, 1 U. S., at 2',
              [case_citation(2, page='2', reporter_found='U. S.', short=True,
-                            metadata={'antecedent_guess': 'Johnson'})]),
+                            metadata={'antecedent_guess': 'Johnson',
+                                      'court': 'scotus'})]),
             # Test short form citation with no comma after reporter
             ('before asdf, 1 U. S. at 2',
              [case_citation(2, page='2', reporter='U.S.',
                             reporter_found='U. S.', short=True,
-                            metadata={'antecedent_guess': 'asdf'})]),
+                            metadata={'antecedent_guess': 'asdf',
+                                      'court': 'scotus'})]),
             # Test short form citation at end of document (issue #1171)
             ('before asdf, 1 U. S. end', []),
             # Test supra citation across line break
@@ -232,12 +237,14 @@ class FindTest(TestCase):
             ('before asdf, 1 U. S., at 20-25',
              [case_citation(2, page='20', reporter_found='U. S.', short=True,
                             metadata={'pin_cite': '20-25',
-                                      'antecedent_guess': 'asdf'})]),
+                                      'antecedent_guess': 'asdf',
+                                      'court': 'scotus'})]),
             # Test short form citation with a page range with weird suffix
             ('before asdf, 1 U. S., at 20-25\\& n. 4',
              [case_citation(2, page='20', reporter_found='U. S.', short=True,
                             metadata={'pin_cite': '20-25',
-                                      'antecedent_guess': 'asdf'})]),
+                                      'antecedent_guess': 'asdf',
+                                      'court': 'scotus'})]),
             # Test first kind of supra citation (standard kind)
             ('before asdf, supra, at 2',
              [supra_citation(2, "supra,",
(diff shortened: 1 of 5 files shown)


mlissner:
@ProbablyFaiz, this is great and we are absolutely accepting outside contributions! We'll be very curious to hear what you're using eyecite for, and we're thrilled to see that it's getting use. 

@jcushman, do you have time/interest/ability to give this one a review? It's short, has tests, and seems like it's pretty good to me, but it builds on your code.

ProbablyFaiz:
> @ProbablyFaiz, this is great and we are absolutely accepting outside contributions! We'll be very curious to hear what you're using eyecite for, and we're thrilled to see that it's getting use.

Awesome, glad to hear it! I've now resolved all the CI issues (enabled the workflows on our fork and dredged out the lint errors) so things should be good to go on that front.

We're working on a project primarily focused on legal citation analysis, so we're using eyecite a ton indirectly by pulling CourtListener data and more recently, directly by using it to pull out citations from a provided document for analysis. This specific PR came from us experimenting with what we can learn from citation metadata, including parentheticals, beyond just the relationship itself. All of this would be literally impossible without all of the FLP's efforts to create and publish all this, so thanks again :smile:

jcushman:
Hi, @ProbablyFaiz! Thanks for working on this, it looks like a great start.

Your current regex is a little too broad -- it will pull in any parenthetical within 150 characters, even if there are other cites in the way:

```
In [3]: get_citations("1 U.S., at 2. blah. 3 U.S. 4 (overruling ...).")[0].parenthetical
Out[3]: 'overruling ...'
```

Looks like full cites also sort of have that problem already, but only for the rare case where we have "3 U.S. 4. blah blah blah (fourdigityear)," whereas pulling in the wrong short cite parenthetical would be common.

To make the regex simpler to nail down, what if rather than adding this to `add_post_citation`, you replace the line where we currently have `pin_cite, span_end = extract_pin_cite(words, index, prefix=cite_token.page)` inside `extract_shortform_citation`? I think if you used the same logic as `extract_pin_cite` but with a regex something like this, it would work:

```
# handle a short cite's extra and parenthetical:
POST_SHORT_CITATION_REGEX = rf"""
    # optional pin cite
    {PIN_CITE_REGEX}?
    \ ?
    # optional parenthetical comment:
    (?:\((?P<parenthetical>[^)]+)\))?
"""
```

(I haven't tested that, but something along those lines.)

You might also throw in a test for this getting-the-wrong-parenthetical case so we don't lose track of it later.

I'm not clear if short cites need an arbitrary `(?P<extra>[^(]*)` the way full cites do. The `<extra>` is for skipping parallel cites like `1 U.S. 1, 2 S.Ct. 2, 3 L.Ed. 3 (1999)`, and I would think it would be rare to have parallel short cites that need to be skipped. But I don't really know. If short cites do need an `<extra>` it has to be narrower than `[^(]*`, so maybe we should leave it out for now until we have examples we're trying to match.

Note that there's a big refactor going on over in #64, so I'm not too worried about style on this one since these changes will get wrapped into the refactor -- just want to get the functionality and tests right.

jcushman:
@ProbablyFaiz if it's easier I'm also happy to add this on top of #64 once that comes in, since a lot of stuff is moving around over there.

ProbablyFaiz:
> @ProbablyFaiz if it's easier I'm also happy to add this on top of #64 once that comes in, since a lot of stuff is moving around over there.

I'll take care of it, no worries. Still debugging regex over here but once I get that figured out I'll rebase on master and update this PR.

ProbablyFaiz:
Alright, this PR is ready for another look. It's gotten a good bit more ambitious in scope than the last version; see the description of the PR for more info.

tests/test_FindTest.py
```diff
@@ -224,12 +249,96 @@ def test_find_citations(self):
...
+                                      'court': 'scotus'})]),
+            # Test short form citation with a parenthetical
+            ('before asdf, 1 U. S., at 2 (overruling xyz)',
+             [case_citation(4, page='2', reporter='U.S.',
+                            reporter_found='U. S.', short=True,
+                            metadata={'antecedent_guess': 'asdf',
+                                      'parenthetical': 'overruling xyz',
+                                      'court': 'scotus'}
+                            )]),
+            # Test short form citation with no space before parenthetical
```
jcushman, line 268: Looks like this test string should have a space removed?

jcushman (approved):
Awesome! Just the one note from me, otherwise this looks good to go.

Great catch on the mistake in the test harness! Whoops. 😬 I'll clean up the indexes once this goes in. I think court is also supposed to be set by the test fixture but was lost in the metadata refactor, so I can fix that too.

mlissner:
Fixes: #67.

tests/test_FindTest.py
ProbablyFaiz: Fixed!

jcushman:
Looks good to me!

mlissner:
This is great! Thank you @ProbablyFaiz for taking this to the finish line, and thank you @jcushman for your review!

PR #63, opened 2021-05-12
Add setuptools to build-system

jcushman:
I don't know if this is actually correct, but if I try to install eyecite in editable mode like "pip install -e /path/to/eyecite", I get "ModuleNotFoundError: No module named 'setuptools'" unless this is included. This way it seems to work.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,4 +72,5 @@ good-names = "i,j,k,ex,Run,_,m"
 
 [build-system]
 build-backend = "poetry.core.masonry.api"
-requires = ["poetry-core>=1.0.0"]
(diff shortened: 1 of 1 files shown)


mlissner:
Well, this is an annoying problem. There's a TON of conversation about this in the poetry issues/discussions/etc. Apparently it has something to do with pypi having a hard requirement on the setup.py file even when the new toml file is available. 

Anyway, this seemed like the most sane solution of the many random things I just read: https://github.com/python-poetry/poetry/discussions/1135#discussioncomment-145763

(I've never used editable mode with pip though. Seems useful. :))

PR #64, opened 2021-05-12
Add extraction of laws.json and journals.json patterns from reporters_db

jcushman:
This adds extraction of laws.json and journals.json patterns from reporters_db:

```
>>> get_citations("1 U.S. 1. Mass. Gen. Laws ch. 1, § 2. 1 Minn. L. Rev. 2.")
[FullCaseCitation('1 U.S. 1', groups={'volume': '1', 'reporter': 'U.S.', 'page': '1'}),
 LawCitation('Mass. Gen. Laws ch. 1, § 2', groups={'reporter': 'Mass. Gen. Laws', 'chapter': '1', 'section': '2'}),
 JournalCitation('1 Minn. L. Rev. 2', groups={'volume': '1', 'reporter': 'Minn. L. Rev.', 'page': '2'})]
```

Yay! Now some questions ...

**Question: class hierarchy?**

Here's the class hierarchy I currently have for citations, with the ones we actually instantiate in bold:

* CitationBase
  * ResourceCitation: stuff done by full and short cites that matched a pattern from reporters_db
    * FullCitation: label for things that are full cites instead of short cites
      * **FullLawCitation**
      * **FullJournalCitation**
    * CaseCitation: stuff done by case citations
      * **ShortCaseCitation**
    * **FullCaseCitation** (inherits from both CaseCitation and FullCitation)
  * **SupraCitation**
  * **IdCitation**
  * **NonopinionCitation**

You can see it gets kind of messy because we need a base class for things that all reporters_db citations do, a base class for full citations, and a base class for case citations, and then the actual models inherit from one or more of those. Dunno if this could be better.

You can also see I'm not currently attempting to deal with short journal or law citations. Seemed like I could get away with saving that for later.

**Question: citation object attributes?**

OK so once you call `get_citations()` and get back some citations, what (public) attributes and methods do they have? This has gotten a little messy. Here's the current answer in this PR, ignoring the generic ones like IdCitation and SupraCitation:

* All:
  * matched_text(): the actual text matched by the regex that found this cite
  * span(): the range of characters in the input text that represent this cite
  * groups: the dict of match groups in the regex that found this cite
  * base_citation(): matched_text() with corrected reporter string based on edition_guess
  * formatted(): full citation with corrected reporter string and metadata extracted from nearby text
  * reporter_found: same as self.groups['reporter']
  * reporter: same as self.groups['reporter'], but if `edition_guess` is set then it becomes `edition_guess.short_name`
  * year: extracted from following text
  * pin_cite: extracted from following text
  * parenthetical: extracted from following text
  * all_editions: list of Edition objects for the regex that found this cite
  * exact_editions: subset of all_editions that had this as their primary regex
  * variation_editions: subset of all_editions that had this as a variation
  * edition_guess: single Edition object, if there's only one in exact_editions with a valid year, or if no exact_editions, only one in variation_editions with a valid year
* FullCaseCitation -- same as All plus:
  * volume: same as self.groups['volume']
  * page: same as self.groups['page']
  * canonical_reporter: if `edition_guess` is set, becomes the name of the cluster from reporters.json, like `F.` for the edition `F.2d`
  * plaintiff: extracted from preceding text
  * defendant: extracted from preceding text
  * extra: extracted from following text
  * court: extracted from following text
* ShortCaseCitation -- same as FullCaseCitation plus:
  * antecedent_guess: word immediately before citation
* FullJournalCitation:
  * volume: same as self.groups['volume']
  * page: same as self.groups['page']
* FullLawCitation:
  * month: extracted from following text
  * day: extracted from following text
  * publisher: extracted from following text

Possible changes to make this less messy:

* We could collect all of the "extracted from following text" variables into a dict named something like `metadata` that includes everything we capture before and after the cite. So remove the attributes `publisher, day, month, antecedent_guess, court, extra, defendant, plaintiff, parenthetical, pin_cite` and just have like `metadata["publisher"]`. That could leave less of a "lots of special cases" feeling.
  * (One special case I would keep -- I'd probably put the raw value we find for year as `metadata["year"]`, but still set `year` as an integer year from a valid range like we do now, because that one seems useful).
* Remove `canonical_reporter`? Not sure if this is doing anything useful, but you can get it from `edition_guess.reporter.short_name`.
* Remove `reporter_found` in favor of `groups['reporter']`?
* Remove `volume` and `page` in favor of `groups['volume']` and `groups['page']`?
* Rename reporter `reporter` to something like `def cleaned_reporter(): return self.edition_guess.short_name if self.edition_guess else self.groups['reporter']`?
* Rename base_citation() and formatted() to cleaned_citation() and cleaned_citation_full()?

If we made all of those changes then the API for all reporters_db citations would become this:

* All:
  * matched_text()
  * span()
  * groups
  * metadata
  * cleaned_reporter()
  * cleaned_citation()
  * cleaned_citation_full()
  * year
  * all_editions
  * exact_editions
  * variation_editions
  * edition_guess

Happy to include something along these lines in this PR, but I didn't want to delete a bunch of stuff if it wasn't going in the right direction.

**Other miscellaneous stuff in this PR:**

* Removed `do_defendant` and `do_post_citation` flags from `get_citations` since those seem harmless to always do and required some extra moving parts to support. In the future there could be some flag for not extracting cite metadata in general?
* Added a `format_resolution` helper function to the resolve_citations tests while getting those working, to make it easier to see why two resolutions don't match.
* Moved all regexes and regex-building helpers to regexes.py, as Mike suggested in the last PR.
* To make it easier to view citation objects in the repl, everything inheriting from CitationBase now uses a simplified `__repr__` like `FullCaseCitation("original matched text", groups={...})`. Some citation classes previously had formatted citation builders in their `__repr__` function, which I think was misnamed -- those now live in a `formatted()` method, which I suggested above might want to be called `cleaned_citation_full()`.

**Stuff left undone:**

* Handling short cites for laws and journals.
* Pin cite extraction for laws is probably pretty incomplete. (And I suspect some stuff might want to move between reporters.json and laws.json once we look more at that -- like maybe anything cited by page range wants to be in reporters.json.)
* Docs need some updates once we settle on the api.
* Just fyi "U.S.C" won't be found yet because I haven't added federal statutes yet over in reporters_db.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -207,11 +207,11 @@ of tokens:
     In [2]: list(default_tokenizer.tokenize("Foo v. Bar, 123 U.S. 456 (2016). Id. at 457."))
     Out[2]:
     ['Foo',
-     StopWordToken(data='v.', stop_word='v'),
+     StopWordToken(data='v.', ...),
      'Bar,',
-     CitationToken(data='123 U.S. 456', volume='123', reporter='U.S.', page='456' ...),
+     CitationToken(data='123 U.S. 456', volume='123', reporter='U.S.', page='456', ...),
      '(2016).',
-     IdToken(data='Id.'),
+     IdToken(data='Id.', ...),
      'at',
      '457.']
 
--- a/eyecite/find_citations.py
+++ b/eyecite/find_citations.py
@@ -1,9 +1,9 @@
-from typing import Iterable, List, cast
+from typing import List, Union, cast
 
 from eyecite.helpers import (
-    SHORT_CITE_ANTECEDENT_REGEX,
-    SUPRA_ANTECEDENT_REGEX,
     add_defendant,
+    add_journal_metadata,
+    add_law_metadata,
     add_post_citation,
     disambiguate_reporters,
     extract_pin_cite,
@@ -14,6 +14,9 @@ from eyecite.models import (
     CitationBase,
     CitationToken,
     FullCaseCitation,
+    FullCitation,
+    FullJournalCitation,
+    FullLawCitation,
     IdCitation,
     IdToken,
     NonopinionCitation,
@@ -23,22 +26,21 @@ from eyecite.models import (
     SupraToken,
     Tokens,
 )
+from eyecite.regexes import SHORT_CITE_ANTECEDENT_REGEX, SUPRA_ANTECEDENT_REGEX
 from eyecite.tokenizers import Tokenizer, default_tokenizer
 
 
 def get_citations(
     plain_text: str,
-    do_post_citation: bool = True,
-    do_defendant: bool = True,
     remove_ambiguous: bool = False,
     tokenizer: Tokenizer = default_tokenizer,
-) -> Iterable[CitationBase]:
+) -> List[CitationBase]:
     """Main function"""
     if plain_text == "eyecite":
         return joke_cite
 
     words, citation_tokens = tokenizer.tokenize(plain_text)
-    citations: List[CitationBase] = []
+    citations = []
 
     for i, token in citation_tokens:
         citation: CitationBase
@@ -53,12 +55,6 @@ def get_citations(
                 citation = extract_shortform_citation(words, i)
             else:
                 citation = extract_full_citation(words, i)
-                if do_post_citation:
-                    add_post_citation(citation, words)
-                if do_defendant:
-                    add_defendant(citation, words)
-            citation.guess_edition()
-            citation.guess_court()
 
         # CASE 2: Citation token is an "Id." or "Ibid." reference.
         # In this case, the citation should simply be to the item cited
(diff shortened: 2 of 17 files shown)


mlissner:
Some general responses, and then I'll go look at the code, since there's plenty here to chew on. 

1. Yes, the API simplification makes sense to me and I vote for it. The one thing that'd be nice is if we used an object for `metadata` instead of a `dict`, so that it can have better inspections and be easier to use. I tire of writing things like `metadata["mything"]` instead of just `metdata.thing` (I guess JS-land is ready for me).

    We do use `canonical_reporter` in a few places in CL, but it'd be easy enough to update our code to get the values we need even if it went away.

2. Doing short cites for journals later seems fine to me. 

3. The class hierarchy is a mess. There's no way I can have that in my head. The one solution that sometimes helps with this is mixins. Maybe there's an opporunity to use them here to simplify, but I don't see that as a blocker for this: Yes, they're a mess, but it's a mess with a purpose and that works.

4. It'd be nice to sequence this after PR #62, so they don't have to update. It's small.

5. All other miscellanea sounds good.

Now to go look at the code.

mlissner:
OK, I've skimmed over the code. No big surprises that I saw except that dang, that regex file is MASSIVE.

jcushman:
OK, I did the API refactor to move citation attributes extracted from text before and after the citation into `cite.metadata`, remove redundant attributes, and provide `corrected_reporter()`, `corrected_citation()`, and `corrected_citation_full()` methods. This means all `ResourceCitation` objects now expose the same API.

> The one thing that'd be nice is if we used an object for metadata instead of a dict

OK, in this update `cite.metadata` is a dataclass with defined fields. `groups` still has to be a dict because we don't know what groups might come back from the regex.

mlissner:
Awesome. No blockers from me. Merging.

PR #68, opened 2021-05-14
Add dump_citations helper

jcushman:
Not sure if this belongs in the library, but when I was adding journals and statutes I found it useful to dump the results for a bunch of CAP cases and eyeball them to see what looked wrong. That resulted in this helper function `dump_citations(cites, text, context_chars=30)`, which has highlighted output that looks like this:

image

This seemed useful enough that I thought I'd throw it in to see if it helps other folks as well -- mostly for us developing the library but maybe also helpful if you're using eyecite for your own purposes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,8 @@
 
 The following changes are not yet released, but are code complete:
 
-...
+Features:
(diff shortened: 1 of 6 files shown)


mlissner:
Nice. Thanks.

PR #78, opened 2021-05-26
Fix Bankr. L. Rep. test for new reporters-db

jcushman:
The latest reporters-db code ([PR](https://github.com/freelawproject/reporters-db/pull/59)) changed the match groups for `Bankr. L. Rep. (CCH) P12,345` so they come in as `groups={"volume": None ...}` instead of not having a volume key. This is because CCH cites sometimes have a volume and sometimes don't, so there's now an optional match group to check for it.

So, this PR just updates the eyecite tests to expect that `"volume": None` value. It will pass against the latest reporters-db but not against the one on PyPI, so I'm not sure how you want to sequence things.

### DIFF
--- a/eyecite/test_factories.py
+++ b/eyecite/test_factories.py
@@ -60,8 +60,7 @@ def case_citation(
         source_text = f"{volume} {reporter} {page}"
     if short:
         metadata.setdefault("pin_cite", page)
-    if volume:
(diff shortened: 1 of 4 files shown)


mlissner:
I just pushed reporters 3.2.0. Want to update for that and we should have tests pass here?

PR #79, opened 2021-05-26
Filter Id. cites by pin cite

### LINKED ISSUE #74: Use pin cite to filter out short cites in clustering
We can improve cite clustering by excluding short cites with implausible pin cites. Example:

"1 U.S. 200. blah blah. 2 We Missed This 20. blah blah. Id. at 22."

We'll currently cluster "Id. at 22" with "1 U.S. 200," but we could refuse on the basis that `int(id_cite.metadata.pin_cite) < int(us_cite.groups['page'])`. And then I think ignore the Id. cite entirely since it must be the product of some error or other.

jcushman:
This filters Id. cites so that pin cites falling outside the 150 pages after the start of the full cite won't match, so we won't make spurious Id. matches for a string like "1 U.S. 100. 1 Missed 25. Id. at 50. 1 Missed 275. Id. at 300."

This requires some arbitrary line drawing since not all cites have pages (like statutes), and not all pin cites are numbers (like "¶ 3"). I arbitrarily chose to say that any Id. can match a cite without a "page" group, and that any cite that doesn't start with a number doesn't match a full cite with a "page" group. These could be improved later when we see what fails.

Also in this PR:

* `models.py`: I tightened up `Resource.__hash__` to not consider full cites identical if they have different `all_editions`, just in case we get two sources in reporters-db that yield the same match groups.
* `test_ResolveTest.py`: I refactored the `resolve_citations` tests to use a new helper, `checkResolution`, that makes it easier to succinctly state how you want a group of citations to resolve. See the help text to `checkResolution` for details. I found this made it a lot easier to look at a test and see what it was testing.

Fixes #74


mlissner:
All seems good to me. @mattdahl, do you want to take a look too since it affects your short form work?

eyecite/resolve.py
```diff
@@ -46,9 +57,41 @@ def _filter_by_matching_antecedent(
     return matches[0] if len(matches) == 1 else None
 
 
+def _invalid_pin_cite(full_cite: FullCitation, id_cite: IdCitation):
```
mattdahl, line 60: Three things:

1. refactor to be `_valid_pin_cite` instead of `_invalid_pin_cite`? These kinds of methods are always more intuitive to me when written "positively." E.g., returning `False` from this method create a confusing (to me) double negative.
2. add `is` prefix, so `_is_invalid_pin_cite`? Again, just improve readability, at least for me.
3. add `-> bool` return signature?

eyecite/resolve.py
```diff
@@ -148,10 +200,10 @@ def resolve_citations(
     resolved.
     """
     # Dict of all citation resolutions
-    resolutions: Dict[ResourceType, List[CitationBase]] = defaultdict(list)
+    resolutions: Resolutions = defaultdict(list)
 
     # Dict mapping full citations to their resolved resources
-    resolved_full_cites: Dict[FullCitation, ResourceType] = {}
+    resolved_full_cites: ResolvedFullCites = []
```
mattdahl, line 206: What's the idea behind switching from a dict to a list of tuples here?

tests/test_ResolveTest.py
```diff
@@ -1,295 +1,283 @@
...
+        # input we're building for self.assertResolution
+        expected_resolution_dict = defaultdict(list)
+        citations = []
+
+        # resources we've found so far
+        resources: List[Resource] = []
+
+        for i, cite_text in expected_resolutions:
+            # extract cite and make sure there's only one:
+            cites = get_citations(cite_text)
```
mattdahl, line 75: This refactor gives me a bit of pause because to test `resolve_citations()` we now must assume that `get_citations()` is working properly. Of course, that's probably a fine assumption, but I've always thought that unit tests should try to atomically test the logic of one method at a time. Just something that stuck out to me, not a deal-breaker.

eyecite/resolve.py
jcushman: The purpose of the dict was to remove duplicates in this block:

```
        if isinstance(citation, FullCitation):
            resolution = resolve_full_citation(citation)
            resolved_full_cites[citation] = resolution
```

... but that wasn't actually removing any duplicates, because identical cites from different parts of the text aren't considered duplicates for hash equality. I got partway through implementing something that would remove duplicates based on semantic equality (and I kept the useful parts of that effort in `models.py`), but then I realized that I was complicating this part of the code for no reason, because duplicates are rare and already need to be handled further down the line anyway, which is why we didn't notice that they weren't being handled here in the first place.

So at that point `resolved_full_cites` was just a list in dict form, and I made it an actual list to make its function more clear.

eyecite/resolve.py
jcushman: I pushed a commit to rename to `_has_invalid_pin_cite` and add the missing return type.

I agree that double negatives are confusing, but I tried `_valid_pin_cite` and found that more confusing, because it also returns true for id cites without pin cites at all, and for full cites without page groups ... all we're really saying here is this id cite doesn't (a) have a pin cite that (b) fits a particular pattern we think is invalid.

tests/test_ResolveTest.py
jcushman: What if instead of thinking of it as an overbroad unit test you think of it as an awesomely readable functional test? :)

I'll say that in practice I find the new thing easier to read/write/debug. E.g. comparing these two, I have an easier time confirming that the second one is testing what the comment says it is testing, and duplicating it to add new tests:

```
full6 = case_citation(reporter="F.2d", metadata={"plaintiff": "Ipsum"})
full9 = case_citation(reporter="F.2d", page="99", metadata={"defendant": "Ipsum"})
short5 = case_citation(reporter="F.2d", metadata={"antecedent_guess": "Ipsum"}, short=True)
...
            # Test resolving a short form citation when its reporter and
            # volume match two possible candidates, and when its antecedent
            # guess also matches multiple possibilities. We expect the short
            # form citation to not be resolved.
            (
                [full6, full9, short5],
                {
                    Resource(full6): [full6],
                    Resource(full9): [full9],
                },
            ),
```

becomes

```
        # Test resolving a short form citation when its reporter and
        # volume match two possible candidates, and when its antecedent
        # guess also matches multiple possibilities. We expect the short
        # form citation to not be resolved.
        self.checkResolution(
            (0, "Foo v. Bar, 1 U.S. 1."),
            (1, "Foo v. Bar, 1 U.S. 2."),
            (None, "Foo, 1 U.S., at 2."),
        )
```

mattdahl:
All fine by me!

mlissner:
Thank you everybody!

PR #81, opened 2021-06-04
v2.2.0 release and bump courts-db

mlissner:
@jcushman could you give this a look? I'm attempting to release version 2.2.0, with the courts-db fix you landed and with all the API reorganization that you did back in #64. 

I think I summarized well, but there was a lot in there and I think a second set of eyes might pick up one something I missed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,9 +5,59 @@
 The following changes are not yet released, but are code complete:
 
 Features:
- - Adds `dump_citations()` to inspect extracted citations.
+ - None yet
+
+Changes:
+ - None yet
+
+Fixes:
+ - None yet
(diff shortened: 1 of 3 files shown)


mlissner:
Here's the rendered readme: https://github.com/freelawproject/eyecite/blob/v2.2.0-release/CHANGES.md

jcushman:
Looking at the diff ( https://github.com/freelawproject/eyecite/compare/v2.1.0...master ), this looks good to me! The one nitpick I found is that the `resolve_citations` function renamed a parameter from `resolve_fullcase_citation` to `resolve_full_citation` to reflect that not all full citations are cases.

PR #82, opened 2021-06-04
More flexible short cite generation

jcushman:
I noticed that we weren't getting short cites for the `Mich.` reporter, which turns out to be because that reporter has a non-standard volume number in its regex, and we currently only look for short cites for the standard regex.

So this PR does three related things:

* Add a `short_cite_re` function to more flexibly generate short-cite regexes from full cite regexes. Currently this is only slightly more flexible than before, but does successfully work for `Mich.`: it tries to insert an " at " between the `(?P<reporter>...)` and `(?P<page>...)` groups. This could get smarter if we figure out rules for other cite formats.
* When generating regexes for variant editions, put them in a single regex like `(?P<reporter>Mich|Mich Foo|etc.)` instead of making a separate regex for each of them. This is a little more efficient, particularly if you aren't using hyperscan, and was convenient to do while fixing up short cites.
* If two regexes have the exact same results, combine them into one, using the new `Token.merge` function. This avoids a situation where we fail to disambiguate properly because two reporters are looking for the same pattern with slightly different regexes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ Changes:
  - None yet
 
 Fixes:
- - None yet
+ - Initial support for finding short cites with non-standard regexes, including fixing short cite extraction for `Mich.`, `N.Y.2d` and `Pa.`. 
 
 ## Current
 
--- a/eyecite/models.py
+++ b/eyecite/models.py
@@ -11,6 +11,7 @@ from typing import (
     Optional,
     Sequence,
     Union,
+    cast,
 )
 
 from eyecite.utils import HashableDict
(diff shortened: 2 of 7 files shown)


jcushman:
Tests are failing because of the dicts-where-lists-should-be we just fixed over in reporters-db.

mlissner:
Just pushed reporters-db 3.2.1 with that fix.

mlissner:
Looks like the toml file isn't happy. Can you include something in CHANGES.md too? (I'm going to try to remember to be better about it.)

mlissner:
It's like clockwork around here today!

Issue #86, opened 2021-08-19
Docs should say that hyperscan is currently linux only, not only x86_64 only.


step21:
During my review for JOSS, I noticed that hyperscan seems to be unwilling to compile on anything non-linux.
This is fine, but in the eyecite readme at least it only says 'x86' which normally means it should at least run on macos x86_64 or/and Windows x86_64, so I think this should be clarified.
openjournals/joss-reviews#3617

jcushman:
Thanks! I agree it should be documented that the python-hyperscan bindings only officially support linux.

Note that in practice HyperscanTokenizer does currently work on MacOS:

```
❯ python3 --version
Python 3.9.6

❯ brew install hyperscan
...

❯ python3 -m pip install hyperscan eyecite
...

❯ python3 -c "from eyecite import get_citations; from eyecite.tokenizers import HyperscanTokenizer; tokenizer = HyperscanTokenizer(); print(get_citations('1 U.S. 1', tokenizer=tokenizer))"
[FullCaseCitation('1 U.S. 1', groups={'volume': '1', 'reporter': 'U.S.', 'page': '1'}, metadata=FullCaseCitation.Metadata(parenthetical=None, pin_cite=None, year=None, court='scotus', plaintiff=None, defendant=None, extra=None))]
```

The underlying hyperscan C++ library is [cross platform](https://intel.github.io/hyperscan/dev-reference/getting_started.html#software), so it should be possible to get python-hyperscan to compile on other platforms as well.

But this doesn't seem to be a tested configuration for python-hyperscan, so it totally makes sense to document what's officially supported.

step21:
Thanks. I didn't try brew, I tried `pip install` which complained about some library so maybe manually installing hyperscan headers would also help. Btw, if you are looking for way to build/test on different architectures I can recommend conda-forge.

step21:
Also, apart from running the included tests, do you have a test dataset you can recommend?

mattdahl:
Hi @step21, Jack added some notes about Hyperscan support here -- https://github.com/freelawproject/eyecite#hyperscan-installation. Does that resolve this issue?

step21:
Yes, thanks.

PR #88, opened 2021-08-19
Document hyperscan x86 installation requirements

jcushman:
As suggested, appropriately enough, in #86.


mlissner:
> appropriately enough, in #86.

👀!

Issue #92, opened 2021-09-16
API/Reference Documentation


cmaimone:
RE: https://github.com/openjournals/joss-reviews/issues/3617

I see the tutorial in the repo readme, but is there reference/API documentation somewhere -- that lists the functions, classes, etc. in the package, with parameters, etc.  (the type generally generated from docstrings)

I expect this type of documentation for packages (or alternatively much more extensive usage guides)

mlissner:
We don't have that at present. I've never generated that. @jcushman do you have any experience with generating this kind of thing?

step21:
This could maybe also be partially automated. There is at least some use of docstrings in the code, these could f.e. be parsed by pdoc, sphinx or a similar documentation tool.

mlissner:
OK, this is now live at: https://freelawproject.github.io/eyecite/, and will auto-generate via a Github Action whenever a new version is tagged and pushed.

Thank you for this request!

step21:
Cool. Looks great.

PR #96, opened 2021-09-19
Generate API docs

mattdahl:
Re #92, this PR uses `pdoc` to generate nice static HTML webpages documenting eyecite's public API. I think that this is a very helpful addition.

In addition to actually adding the documentation, I cleaned up our API a bit:
- 3a783f8 and c78a9f3 rename the `clean` and `find` module to be consistent with `resolve` and `annotate`.
- 7372d0d moves the `clean_text()` method from `utils.py` into `clean.py`, as I saw no reason to put it over there. This will not affect anyone currently doing `from eyecite import clean_text`.
- 23e0c09 renames `annotate()` to `annotate_citations()` to be consistent with `find_citations()` and `resolve_citations()`. This was also necessary because `pdoc` didn't like the name collision between `annotate` the module and `annotate` the function. Unfortunately, this does break external imports.
- ac476b4 removes `dump_citatations()` from being exposed on the top `eyecite` module. I saw no need for this, since it's just a debugging function. People can import `eyecite.utils.dump_citations` if they want.

I also wrote a short script, `api_documentation/generate_documentation.sh`, to automate the creation of the documentation. One problem was that `pdoc` insisted on evaluating and printing the full tokenizer definition (thousands of lines of text) in the `get_citations()` definition (since we instantiate a default tokenizer there), and I couldn't figure out how to suppress this, so I just used `sed` to delete that portion from the docs after they were created.

Do we want to serve these somehow? One could in theory browse them locally (starting at `index.html`), but it could be nice to link to them.

Corrections/improvements welcome.


mlissner:
At a glance this seems overall good to me. Have you done any research on
what it takes to get this to readthedocs or similar?

Le sam. 18 sept. 2021 à 18:52, Matt Dahl ***@***.***> a
écrit :

mattdahl:
I did not, though I should have in hindsight. Infuriatingly, it seems that readthedocs does not support `pdoc`. I can try to regenerate this using sphinx instead, or we can "self-host" it using GitHub pages. I believe the URL would be `https://eyecite.github.io`. `pdoc` hosts itself that way, for example: https://pdoc3.github.io/pdoc/

mattdahl:
Actually, I think it would be `https://freelawproject.github.io/eyecite`

mlissner:
I don't have strong feelings on it, personally. I guess the biggest criterion for me is whichever one is the easiest to automate. I have some familiarity with GH pages, so I guess I lean towards learning one fewer thing by going that way? But lots of people use Read The Docs, so I assume it has something really compelling going for it? Bit of a rambly reply, sorry.

mattdahl:
I actually like GH pages too. I think it should be pretty easy to automate. It seems that we can configure GitHub to serve static files from a particular folder (what I've called `api_documentation` or anything else) -- https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#choosing-a-publishing-source. Then anytime we push new files to that folder on `main` the website should automatically update. The only manual thing would be re-running `pdoc` (and my `sed` hack) if any of our docstrings change, though in theory that could also be automated with an Action, I would think.

mlissner:
I'm reviewing all your work here. Doing a GH page from a folder makes sense. It'd be great if it was generated automatically with an Action. Would you want to try your hand at that and include that in this PR?

mattdahl:
I have no idea if this will work, but I just took a shot at it. I've never created an Action before. The idea is that when somebody opens a PR, `pdoc` gets run and any changes to the docs are committed for inclusion in the PR.

As I was typing this, I see my changes already managed to break all the checks. Not sure why that would be -- aren't Actions self-contained/siloed? Why would adding a new one break an old one? I don't really know what I'm doing here.

mlissner:
Matt, thanks for getting the bulk of this figured out. I did a bit of work making the dumb github action work properly, and then tweaked the change log. I think it should do the auto-build now, when I merge it, but only when somebody merges it. We'll find out in a sec! 

Anyway, I'll keep poking at it. We should update links to point to the new docs too, once the gh page is working.

mlissner:
This is live and looks great: https://freelawproject.github.io/eyecite/

mattdahl:
No, thank you! I had no idea how to set up the environment correctly for the action...

Issue #95, opened 2021-09-16
References to "database" in tutorial


cmaimone:
RE: https://github.com/openjournals/joss-reviews/issues/3617

There are references to a "database" in the tutorial - I don't know what these mean.  What database?  Are you talking about linking to external legal databases of some type?  Are these just adding URLs?

It also says things like "fetched from our database" -- but what is "our database"?

cmaimone:
Related to the above, the entire section

> Now we want to resolve all of the extracted cites into clusters indexed by the resource they refer to, such as a case or statute. We'll use a custom function to resolve a given full cite to its resource, so we can return our own MyCaseModel for citations we recognize. We'll fall back on returning resolve_full_citation() for citations we don't recognize.
> 
> For this simplified example, we'll assume we have a database model MyCaseModel so that MyCaseModel.get_by_citation() will return the case referred to by that citation string. In real life this might be a Django model or Elasticsearch lookup. We'll also assume that the same case has the parallel citations "1 U.S. 2" and "1 S. Ct. 2", so MyCaseModel.get_by_citation("1 U.S. 2") returns the same case as MyCaseModel.get_by_citation("1 S. Ct. 2").

I got lost.  I think you're saying you have a hypothetical database with a table model class in it, separate from this package.  This is super confusing for folks who aren't familiar with this type of setup or don't have their own database.  Do you have to use a database for the citation resolution?  All code in the tutorial (with maybe a few exceptions for special use cases, etc.) should be able to run by a potential user.  I can't test the resolve_citations() functionality because the example code uses a database I don't have (and there aren't instructions to set one up.

mlissner:
Hopefully this should be clear now too. Could you take a look and hopefully it'll be better? Thank you for the helpful feedback.

cmaimone:
It's clearer, but I still think the code in the tutorial should be mostly run-able if possible.  For the custom_resolution_callback function, could you make a very simple one that just looked up values in a dict or something similar?  And also code to make your SomeCustomObject class?  This "class" gets used later for annotation examples too.  

Or can you call one of the APIs if they're open (no key) to show a real example?

I'm not going to hold up the paper publication over this, but sorting out some of these things should help eyecite be more clearly and functionally a stand-alone package, that doesn't rely on such close integration with the systems it was developed for.

mattdahl:
[...]

PR #100, opened 2021-09-26
More tutorial improvements re #95

mattdahl:
This PR transforms our tutorial into a Jupyter notebook, which we can link to from the README and should render pleasantly from within GitHub. Re #95, this change should (1) provide an example of how to use `eyecite` with a real document containing thousands of real citations and (2) make clear that all of the tutorial code is run-able by the reader.

The pseudocode about passing a custom resolution to `resolve_citations()` has been removed entirely, as we already explain the same idea here: https://github.com/freelawproject/eyecite#resolving-citations.


mlissner:
I don't know if you still have energy for this, so just merging since it's another nice improvement, but I had a few questions when reading through this:

 - We mention that the hyperscan regexes are pre-compiled, but never explain how to do that.

 - We frequently say something like, "We need to clean the text, so we run `clean(text, [html, whitespace])`. These kinds of statements would be better if we explain what the arguments are that we're passing to `clean()`. Something like, "Clean takes two arguments. The first is the text to be cleaned, and the second is an array of cleaning utilities to run. We have several built in utilities, such as...."

 - The section around resolving citations is a bit unclear. This bit caught me up: "This opinion contains more than 1000 raw citations, but that doesn't mean that this opinion cites 1000 different cases." Maybe say something like, "This opinion contains more than 1000 citations, but these are not all full citations like "123 XYZ 456". In addition to these more obvious citations, `eyecite` will also find short-form citations such as "id" and "supra". So, while there are 1005 citations total, the count of unique opinions cited is much fewer."

Anyway, my policy on documentation is to merge any form of progress, so this is merged, but those were the spots I got caught. Is it easy to edit a Python notebook?

mattdahl:
I still have some energy lol, and I agree that making the second and third change would be good. For the first one, it's my understanding that the pre-compiling happens automatically, right? I.e., the first time the tokenizer is instantiated, it pre-compiles all the regexes and dumps them into the cache folder. Then subsequent calls to `get_citations()` just use that without having to re-compile anything. I can state this explicitly, but there's nothing the user has to do, right?

mlissner:
Yeah, I think that's right, but I haven't witnessed it myself yet. If so, we should state it though. Maybe @jcushman can confirm.

Issue #84, opened 2021-06-26
Don't set court='scotus' for South Carolina citations


mattdahl:
Eyecite thinks that South Carolina citations are SCOTUS citations:

```
from eyecite import get_citations
text = 'Lee County School Dist. No. 1 v. Gardner,  263 F.Supp. 26 (SC 1967)'
cites = get_citations(text)
cites[0].metadata.court

# prints 'scotus'
```

The `SC` in the year could be ambiguous, but the `F.Supp.` reporter should automatically rule SCOTUS out as a possibility for the court here.

devlux76:
You will never see SC in the year field of a Supreme Court decision (if the person who wrote it is citing things properly).
The court is only included when the relevant court is unclear from the reporter cited. 
The Supreme court will always be cited to the U.S. or the S.Ct. reporters unless it's a slip opininon, so there should never be any ambiguity.

**Bluebook R. 10.4(b) State courts.**
>*In general, indicate the state and court of decision. However, do not include the name of the court if the court of decision is the highest court of the state.* 
The Bluebook: A Uniform System of Citation R. 10.4(b), at 106 (Columbia L. Rev. Ass’n et al. eds., 21st ed. 2020).*

mlissner:
Thanks @devlux76. This looks like a great first bug. Any interest in trying to tackle it with a test and a fix?

devlux76:
Sure!  Do we know where this citation came from?
A search of "263 F.Supp. 26" on Westlaw shows just 5 citing cases.

[Bowen v. Massachusetts, 487 U.S. 879, 108 S. Ct. 2722, 101 L. Ed. 2d 749 (1988)](https://www.courtlistener.com/opinion/112143/bowen-v-massachusetts/)

[Bd. of Pub. Instruction of Palm Beach Cty., Fla. v. Cohen, 413 F.2d 1201 (5th Cir. 1969)](https://www.courtlistener.com/opinion/285866/board-of-public-instruction-of-palm-beach-county-florida-v-wilbur-j)

[Fort Sumter Tours, Inc. v. Andrus, 440 F. Supp. 914 (D.S.C.), aff'd, 564 F.2d 1119 (4th Cir. 1977)](https://www.courtlistener.com/opinion/1817921/fort-sumter-tours-inc-v-andrus)

[...]

PR #105, opened 2021-12-25
 Don't set court='scotus' for South Carolina citations #84 

devlux76:
This addresses issue #84. The issue was caused because SC is treated differently than S.C. when searching and scotus is the default return result. Since all court strings are unique with or without punctuation, stripping all punctuation and converting to the same case allows for an exact comparison instead of the "startswith" comparison that was previously being used.


mlissner:
@jcushman do you have any bandwidth or interest in taking a look at this one?

@devlux76, thanks! Looks like you need to do a little linting work now that tests have completed though.

devlux76:
@mlissner I'm sorry, I don't understand. 
>Looks like you need to do a little linting work now that tests have completed though.

Can you please give me more info about what needs to be done here?
Thanks!

mlissner:
Linters are programs that check your work to make sure it's clean and pretty and using best practices. We have a few of them that were in Github Actions, but we didn't have a good process for developers to actually run them, so I just added a dev dependency for pre-comit to this branch. We use this in most of our other projects, so it seems reasonable to use it here too. 

After adding it, I ran it, it cleaned up and highlighted a few problems with your code, which I fixed for you.

So this is ready for review again, but I'll wait for Jack to see if he wants to look at it. If not, I'll jump back in.

devlux76:
Ok, I understand on the linting. I guess I'm so used to golint autoliniting my code that I forgot python doesn't do this automatically.

So for future reference, is there a particular process I should use prior to committing?

mlissner:
If you run `pre-commit install` in your repo, it'll install a git hook that runs the pre-commit checks before each commit. If it fails, some things (like formatting) will be fixed automatically, and you can just re-run your git commit. Other things will report errors you'll have to fix yourself. But that's the heart of it, and you'll write better code as a result.

PR #108, opened 2022-01-26
backtrack to find full plaintiff

neilfulwiler:
untested, this is just an idea about how this might work. And then I'm going to figure out how I can incorporate this into other python projects without pushing to pip registries, which I don't really know how to do


neilfulwiler:
sorry this wasn't intended to go here :grimacing:

Issue #109, opened 2022-02-01
Rename NonopinionCitation to UnknownCitation


mattdahl:
`NonopinionCitation` seems like a misnomer now that we have `FullLawCitation` and `FullJournalCitation` abstractions (which are obviously not opinions). The point of `NonopinionCitation` is to serve as a naive catch-all for any citation that can't be otherwise parsed, so I think it should be renamed to something like `UnknownCitation`.

mlissner:
Makes sense. Ah to have a stable API. Someday!

mattdahl:
Uh, so it seems like you agree, but you also closed this issue? Not sure how to interpret :)

mlissner:
Oops, didn't mean to.

mattdahl:
Fixed in #110

PR #110, opened 2022-02-06
Rename NonopinionCitation to UnknownCitation

mattdahl:
Addresses #109.

### DIFF
--- a/eyecite/test_factories.py
+++ b/eyecite/test_factories.py
@@ -6,11 +6,11 @@ from eyecite.models import (
     FullLawCitation,
     IdCitation,
     IdToken,
-    NonopinionCitation,
     SectionToken,
     ShortCaseCitation,
     SupraCitation,
     SupraToken,
+    UnknownCitation,
 )
 from eyecite.tokenizers import EDITIONS_LOOKUP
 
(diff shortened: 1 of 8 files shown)


mlissner:
Is it worth it to do an alias here and emit a deprecation warning? We haven't done that in the past, but it'd be pretty easy to do, I think.

mattdahl:
Interesting idea -- I'm not totally sure on the pattern here, but I just pushed a commit that takes a stab at that. I said it would be deprecated in 2.5.0 on the logic that the next release will be 2.4.0, so it should be removed in the release after that one?

eyecite/test_factories.py
```diff
@@ -98,11 +98,20 @@ def id_citation(source_text=None, index=0, **kwargs):
...
+
+def nonopinion_citation(*args, **kwargs):
+    from warnings import warn
+
+    warn(
+        """nonopinion_citation() will be deprecated in eyecite 2.5.0.
+        Please use unknown_citation() instead.""",
+        DeprecationWarning,
     )
+    return unknown_citation(*args, **kwargs)
```
mlissner, line 114: I don't think we need deprecation warnings in tests, but otherwise, this PR looks ready to me.

mattdahl: This isn't in our tests per se -- the test factories are part of our public API that other people can use to write _their_ tests. So if we're going to alias the class I think we should alias this factory function too.

mlissner: OK, yeah, that *does* make sense. Merging!

mlissner:
Thank you Matt!

Issue #30, opened 2021-03-02
Find partial citations for decisions that don't yet have full citations, e.g. 22 U.S. ____ (2002)


mlissner:
Back in 2014, in https://github.com/freelawproject/courtlistener/issues/1601, @brianwc lamented that we don't handle partial citations very well. His example was:

> The Supreme Court often cites its recent cases that don't yet have a full U.S. citation like so:
> 
> Bullcoming v. New Mexico, 564 U. S. ___ (2011).
> 
> We have Bullcoming v. New Mexico, know it is a SCOTUS case, and know it was filed in 2011. 

Alas, we never got the citator working on these. I'm not sure it's worth the effort to do so now, but I thought I'd file an issue so we could track it as a gap in functionality.

mattdahl:
Relinking my old PR on this: https://github.com/freelawproject/courtlistener/pull/1226

On the parsing side, I think this may be as simple as allowing "__" to be a valid page number here: https://github.com/freelawproject/eyecite/blob/master/eyecite/utils.py#L31

mlissner:
Yeah, I think we should start by focusing on the parsing side first. On the matching side, we have competing proposals of either:

1. Building up a list of all possible volumes for all reporters; or
2. Just doing lookups based on ± 1-2 years from the date of the opinion.

I actually think number two is more practical and less fragile. I'd say if we revisit this once the parser is done, that we take that approach.

devlux76:
I'm volunteering to pitch in on a solution to this if you'll have me.
I'm a 2YL, just finished up my "electronic legal research" and "legal analysis & writing a.k.a how to cite to bluebook" courses. I got decent grades. I'm also a moderately competent coder. I've got some spare time on my hands before it's back to law school.

I have some ideas I'd like to lay down. Just don't want to step on anyone's toes.

mlissner:
Go, go, go, @devlux76. We'd love the help. I think I highlighted one other issue that would be a good first-bug, but this would be a good one too, I think.

flooie:
@mlissner would you like to revive this?

mlissner:
T shirt size estimate?

Issue #76, opened 2021-05-24
Link parallel cites


jcushman:
For a cite like "1 U.S. 1, 2 S. Ct. 2 (1999) (overruling ...)" we extract "1 U.S. 1" and "2 S. Ct. 2" as separate cites that both have the parenthetical "overruling ...". If you later report the parentheticals somehow you double up, or if you use a resolver that knows those are the same case, you double-count the weight of that citation. It would be good if we detected this and linked the two cites as parallel to each other, so the weight and parenthetical could only be counted once.

mlissner:
We have a little (old, unfinished) code related to this here. It has a magic variable identifying the distance between two valid parallel citations, but I forget how carefully I analyzed that:

https://github.com/freelawproject/courtlistener/blob/3f038ff702050a193626bfba7372acf1fa167580/cl/citations/tasks.py#L19-L60

We also have a script that I never quite finished that was designed to find parallel citations (using the task above), build them into a weighted graph and then identify valid parallel citations according to the weight of the graph edges: 

https://github.com/freelawproject/courtlistener/blob/2b6cae362b86e73201bb09be2c133653ca5b9d42/cl/citations/management/commands/cl_add_parallel_citations.py#L254-L270

Doing this upstream in eyecite makes sense to me.

jcushman:
Quick answer with the brain I have available ... :)

**Data representation**

I like Mike's second approach fine. I think the first one centers an edge case too much, you shouldn't have to wade through that layer when most cites aren't parallel. A third option that I would strongly consider would be to only include the first cite in the output list, and attach the parallel cites to that:

```
[
  {reporter, page, volume, parenthetical, parallel_cites: [{reporter, page, volume}]}
]
```

This has the benefit that if you're a naive caller of the library who has no idea about parallel cites, you'll do a reasonable default thing of ignoring them instead of double counting anything. I'm not sure what data exactly belongs in `parallel_cites` but hopefully the answer would jump out during implementation.

(Note that besides doubled parentheticals, there's a couple of other weird things about the current situation, like the first site has a `pin_cite` that contains the second cite, and the second site has a title that contains the first cite. So throwing away the second cite entirely rather than just tagging the second one as special somehow is appealing.)

**Implementation**

[...]

PR #116, opened 2022-07-16
Issue 30 find partial page citations

mattdahl:
This enables `eyecite` to find citations that don't yet have a final page number, but just have a placeholder `___` space instead. This is common with the `U.S.` reporter. This simple parsing change is the first half of addressing #30; figuring out how to match these partial citations will have to be done in CL.

Two things I thought of but didn't include:
1) Should we add a metadata flag to the citation objects like `has_proper_page_number` or something? This would be `False` if the page number is all underscores, `True` otherwise. I wasn't sure if this would actually be helpful or not.
2) Are there other characters besides underscores that courts use to indicate a placeholder for an eventual page number?

### DIFF
--- a/tests/test_FindTest.py
+++ b/tests/test_FindTest.py
@@ -181,8 +181,11 @@ class FindTest(TestCase):
             # Test with page range with a weird suffix
             ('559 N.W.2d 826|N.D.',
              [case_citation(page='826', reporter='N.W.2d', volume='559')]),
-            # Test with malformed/missing page number
+            # Test with malformed page number
             ('1 U.S. f24601', []),
+            # Test with page number that is indicated as missing
+            ('1 U.S. ___',
+             [case_citation(volume='1', reporter='U.S.', page='___')]),
             # Test with the 'digit-REPORTER-digit' corner-case formatting
             ('2007-NMCERT-008',
              [case_citation(source_text='2007-NMCERT-008', page='008',
--- a/tests/test_ResolveTest.py
+++ b/tests/test_ResolveTest.py
@@ -241,6 +241,14 @@ class ResolveTest(TestCase):
             (1, "Ala. Code § 92"),
             (1, "Id. at 2000"),
         )
+        # Test resolving an Id. citation with a pin cite when the previous
+        # citation only has a placeholder page. We expect this to still resolve
+        # because it's possible to pin cite to an opinion with pending final
+        # page numbers.
+        self.checkResolution(
+            (0, "Foo v. Bar, 1 U.S. ___"),
+            (0, "Id. at 100."),
+        )
 
     def test_non_case_resolution(self):
         """Test law and journal resolution."""
--- a/eyecite/regexes.py
+++ b/eyecite/regexes.py
@@ -63,7 +63,8 @@ ROMAN_NUMERAL_REGEX = "|".join(
 # (ordered in descending order of likelihood)
 # 1) A plain digit. E.g. "123"
 # 2) A roman numeral.
-PAGE_NUMBER_REGEX = rf"(?:\d+|{ROMAN_NUMERAL_REGEX})"
+# 3) A page placeholder. E.g. "Carpenter v. United States, 585 U.S. ___ (2018)"
+PAGE_NUMBER_REGEX = rf"(?:\d+|{ROMAN_NUMERAL_REGEX}|_+)"
 
 # Regex to match punctuation around volume numbers and stopwords.
 # This could potentially be more precise.


tests/test_ResolveTest.py
```diff
@@ -241,6 +241,14 @@ def test_id_resolution(self):
             (1, "Ala. Code § 92"),
             (1, "Id. at 2000"),
         )
+        # Test resolving an Id. citation with a pin cite when the previous
+        # citation only has a placeholder page. We expect this to still resolve
+        # because it's possible to pin cite to an opinion with pending final
```
mlissner, line 246: Really? You can pin site using a page before you have a citation with the page? That's truly weird.

tests/test_FindTest.py
```diff
@@ -181,8 +181,11 @@ def test_find_citations(self):
             # Test with page range with a weird suffix
             ('559 N.W.2d 826|N.D.',
              [case_citation(page='826', reporter='N.W.2d', volume='559')]),
-            # Test with malformed/missing page number
+            # Test with malformed page number
             ('1 U.S. f24601', []),
+            # Test with page number that is indicated as missing
+            ('1 U.S. ___',
+             [case_citation(volume='1', reporter='U.S.', page='___')]),
```
mlissner, line 188: Do we use None for the `page` attribute in any other context? Maybe it's better to normalize these underscore things to `None` and then that answers your question about whether we need to add an extra attribute to the response to indicate this situation. What do you think?

mlissner (changes requested):
Wow, this looks simple enough, thanks Matt! 

@jcushman do you have time to take a look and chime in before we merge? 

(A couple questions/comments in the review.)

tests/test_ResolveTest.py
mattdahl: Hmm, maybe not actually, I thought that people would just pin cite to whatever the document page number is (starting at 1) if the real page number was not available, but now I'm not sure. Is it better to expect this test to not resolve the Id. citation?

tests/test_FindTest.py
mattdahl: That's a good idea, I don't think we use `None` in the `page` attribute at all yet!

jcushman:
> @jcushman do you have time to take a look and chime in before we merge?

Not with anything well informed. :)

For Id. cites I imagine they tend to also give a parallel cite with page numbers, e.g. S.Ct. or a docket number, and you could cite to that? Like "Foo v. Bar, 1 S.Ct. 3, 4 U.S. __ (2000). ... Id. at (something)". I'm not sure how that's handled, though. It wouldn't be a terrible idea to grep through for 2 or 3 underscores and see how these are actually used. Possibly one answer is there's always a parallel cite that's better for resolution of the citation.

tests/test_FindTest.py
mattdahl: Fixed in a120ca8

tests/test_ResolveTest.py
mattdahl: Fixed in 6048422

mlissner:
I see you working, but let me know when this is ready for another review.

mattdahl:
> For Id. cites I imagine they tend to also give a parallel cite with page numbers, e.g. S.Ct. or a docket number, and you could cite to that? Like "Foo v. Bar, 1 S.Ct. 3, 4 U.S. __ (2000). ... Id. at (something)". I'm not sure how that's handled, though. It wouldn't be a terrible idea to grep through for 2 or 3 underscores and see how these are actually used. Possibly one answer is there's always a parallel cite that's better for resolution of the citation.

I found this to be sometimes the case but often the Id. cite is just `Id., at ___`, so I changed our behavior to just not match those. Enhancing this with the parallel cite info could be done as a part of #76 perhaps.

> Sounds good to me.

Hashability changes done in 08b9625.

----

N.B. I also had to make a change to the joke cite (5ee4145) because it was causing an issue with the `CitationBase.__post_init__` method for some reason after I implemented the missing page normalization logic.

mlissner (commented):
A couple little things, but looks close.

eyecite/models.py
```diff
@@ -83,6 +83,12 @@ def __post_init__(self):
             if isinstance(self.metadata, dict)
             else self.Metadata()
         )
+        # Set known missing page numbers to None
+        if (
+            "page" in self.groups
+            and self.groups["page"] == len(self.groups["page"]) * "_"
+        ):
+            self.groups["page"] = None
```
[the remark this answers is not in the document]
mattdahl: Good point, I thought my way was faster, but I just tested it and the difference is negligible. Fixed in [d9966ce](https://github.com/freelawproject/eyecite/pull/116/commits/d9966cea0438f149b547c43cff19b4f5f27e2e15)

mlissner:
very nice, thank you!

PR #118, opened 2022-08-09
Update README.rst to reflect name of annotate_citations function

bbernicker:
Current examples raise: `TypeError: 'module' object is not callable`. It looks like the name of the function was changed to annotate_citations and the README was not updated.


mlissner:
Thanks. Sorry about that.

PR #120, opened 2022-08-11
Benchmarking PR

flooie:
This is a Draft with expected updates, but I want to validate the main guts of this PR before I lost any more time to this.

This should in theory generate a chart and time comparing this branch against the main branch and display a report of updates.  But As this branch is nearly identical to the main branch for reasons, it may not show a difference.


flooie:
Actually, for purposes of testing, @mlissner I adjusted the poetry lock file to require an older version of reporters-db and you can see the difference.  

@jcushman I ran with your suggestion. 

There remains a local.sh script to let you run your current branch against the remote/main and generate an output for speed and improved/unimproved results

flooie:
This PR should recreate the previous PR so we dont get a dozen images/charts on a PR with a few tweaks.  

In the case of a PR the GitHub action will compare the current branch vs. the main branch.  I havent capped the table length yet, but I think it will become evident if that is needed over time.   Or in the next big push by @quevon24

jcushman:
Very cool!

I didn't see, is there a length cap in the output chart, if a PR happens to generate 100,000 differences or something?

flooie:
@jcushman no - there isn't a length cap - but I plan to add one - once I got a general consensus ... or got feedback about it.

flooie:
Something, or some GitHub action auto committed the artifacts into the home directory.  Testing now to re-confirm.  Apologies in advance.

.github/workflows/pr_report.yml
```diff
@@ -0,0 +1,44 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: snok/install-poetry@v1
+        with:
+          version: 1.1.13
+          virtualenvs-create: true
+          virtualenvs-in-project: true
+
+      - name: Install Poetry and run Bash script
```
mlissner, line 17: This name is wrong.

flooie: the install poetry?  part.  Install dependencies in poetry and run benchmarks.?

pyproject.toml
```diff
@@ -30,7 +30,7 @@ include = ["eyecite/py.typed"]
 
 [tool.poetry.dependencies]
 python = "^3.7"
-reporters-db = "^3.2.2"
+reporters-db = "3.2.2"
```
flooie, line 33: This was added for testing purposes and will be reverted to the main when complete.

flooie:
@jcushman @mlissner - Ive truncated changes at 50 for the purposes of the PR comment.

flooie:
And ... the PR should've evened out to the same thing with no changes.  Checking my local run to figure out the reason.

mlissner:
Sorry if I missed it. Is this guy ready for review?

flooie:
Yes.  Sent from my iPhoneOn Aug 12, 2022, at 8:19 PM, Mike Lissner ***@***.***> wrote:﻿
Sorry if I missed it. Is this guy ready for review?

benchmark/benchmark.md
```diff
@@ -0,0 +1,34 @@
...
+    sh benchmark.sh -m main -b the-branch-for-the-pr
+
+if you wish to run this test against your local repository vs head/main
+
+    sh benchmarck.sh -l true (or any value) 
+
+Additionally if you want to run the larger corpus for any of these
+tests you can run the flag `-s` (size)
+    
+    sh benchmarck.sh -l true -s true
```
mlissner, line 23: These funky `true` values are another strike against bash. I think we could do these better, but we lack the expertise. :/

benchmark/benchmark.md
```diff
@@ -0,0 +1,34 @@
...
+
+this will use the bulk data ten-percent file 
+
+
+## Reports
+
+The following commands will download or checkout various branches and
```
mlissner, line 30: Is `following` still what you mean here?

benchmark/benchmark.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def fetch_citations(self, row) -> None:
+        """Fetch citations from rows opinion data
+
+        return: None
+        """
+        row_id = row[0]
+        row = dict(zip(self.fields, row))
+        non_empty_rows = [
+            row[field] for field in self.fields if type(row[field]) == str
```
mlissner, line 35: Not sure what's happening here. A comment, perhaps?

benchmark/benchmark.py
```diff
@@ -0,0 +1,78 @@
...
+        self.totals.append(self.count)
+        self.times.append((datetime.datetime.now() - self.now).total_seconds())
+
+    def generate_branch_report(self, branch: str) -> None:
+        """Generate Branch v Main Report
+
+        return: None
+        """
+        zipfile = bz2.BZ2File(
+            Path.joinpath(self.root, "..", "bulk-file.csv.bz2")
```
mlissner, line 58: The location of this file feels a bit magic. Like, it'd be nice if running benchmark.py worked even if benchmark.sh hadn't been run, but since downloading happens elsewhere, we just bake those assumptions in here.

benchmark/chart.py
```diff
@@ -0,0 +1,136 @@
...
+    Generates (mostly) the markdown report for the PR comment
+
+    Returns: None
+    """
+    gains = []
+    losses = []
+    dfA = pd.read_csv(Path.joinpath(root, "outputs", f"data-{branch1}.csv"))
+    dfB = pd.read_csv(Path.joinpath(root, "outputs", f"data-{branch2}.csv"))
+
+    # Make the lists equivalent in length
```
mlissner, line 28: Why is this needed? Maybe a comment?

benchmark/chart.py
```diff
@@ -0,0 +1,136 @@
...
+    # Remove columns to enable comparisons
+    del dfA["Time"]
+    del dfB["Time"]
+    del dfA["Total"]
+    del dfB["Total"]
+
+    comparison = dfA.compare(dfB)
+
+    with open("outputs/output.csv", "w") as f:
+        writer = csv.writer(f)
```
mlissner, line 43: A dictwriter would be better here too.

benchmark/chart.py
```diff
@@ -0,0 +1,136 @@
...
+        writer.writerow(["ID", "GAIN", "LOSS", "OPINION_ID", "--"])
+        for row in comparison.iterrows():
+            non_overlap = set(ast.literal_eval(row[1][0])) ^ set(
+                ast.literal_eval(row[1][1])
+            )
```
mlissner, line 48: Woah, fancy stuff and a scary eval. A comment please?

benchmark/chart.py
```diff
@@ -0,0 +1,136 @@
...
+
+import matplotlib.pyplot as plt
+import pandas as pd
+
+csv.field_size_limit(sys.maxsize)
+
+root = Path(__file__).parent.absolute()
+
+
+def compare_dataframes(branch1: str, branch2: str) -> None:
```
mlissner, line 16: Can this be broken into more functions?

benchmark/chart.py
```diff
@@ -0,0 +1,136 @@
...
+        f.write(
+            "The following chart illustrates the gains and losses "
+            "(if any) from the current pr.\n"
+        )
+
+    # Add markdown report file outputs
+    df = pd.read_csv("outputs/output.csv")
+
+    with open("outputs/report.md", "a") as md:
+        if df.__len__() > 51:
```
mlissner, line 79: Let's put 51 into a variable at least. No magic numbers. Notionally: `MAX_ROWS_IN_MD`?

mlissner (changes requested):
Cool. I've got mostly general comments at this point, then a handful in the chart.py file.

1. pr_report.yml should probably be renamed. Maybe benchmark.py?

1. I'm finding it a bit difficult how things are broken up between the bash script, the benchmark script, and the chart script. I see the appeal of the bash script, since it's running a handful of shell commands, but I think most of those commands could be done pretty easily in Python (creating dirs is easy, downloading files, same; doing git stuff isn't super easy, but it's the exception?). 

    I know you already rewrote it in bash instead of as a Makefile, but I have to admit I can't make much sense of it and it seems like it has a lot of repeated, finicky code. 

    Tell me if you disagree or if you're tired of redoing it, but I'll give a few examples of what I mean:

    Example 1. This makes me nervous, but I can't really explain why. Just feels like there's a good chance something is going to get nuked (I also forget what `-d` does, which, yes, is on me, but also, on bash; is it checking if a directory exists?):
    ```
    [ -d "eyecite/" ] && rm -rf eyecite/
    ```

    Example 2. I'm really not sure what this is doing. I understand all the pieces, but I'm having a hard time synthesizing it all and making sense of it.
    ```
    cd "$PWD/../" && git clone "file:///$PWD" "$PWD/benchmark/eyecite/"
    ```

    Example 3. Am I wrong that you download the sample file twice or maybe three times? You do this on L21-22:

    ```
    [ -d "eyecite/" ] && rm -rf eyecite/
    curl $url --output bulk-file.csv.bz2
    ```
   Then on L33:
    ```
    curl $url --output bulk-file.csv.bz2
    ```

    I can't help but think this is because doing simple things in bash is hard, so mistakes creep in.

    Example 4. There's just a lot of installing dependencies and poetry?

    I don't know. How do you feel about this code?

4. Do you need the `__init__.py` file? If all you want to do is make sure the directory exists, I think I'd avoid making it a python module and instead just put a `placeholder.txt` file in there. In the placeholder file you can explain that the file is needed to keep the directory around. I do wonder though if we need the directory around at all? 

5. Should we add some tweaks to .gitignore for the output directory?

6. Does this use the HyperscanTokenizer parser or the regular one? It should use the Hyperscan if it can, since that's the one that's usually used in prod (and it's faster).

7. We should be using the cleaning tools before processing the data to strip spaces/html/etc.

8. In benchmark.py, there's funky stuff happening with `zip()`, and self.fields. Why not use a DictReader? Similarly, why create arrays of data that need to be zipped back together instead of creating one array of dicts or objects that have the fields you need? Feels like the data isn't shaped properly for how you are creating it or using it.

9. In benchmark.py you establish self.now when starting initializing the class. Would it be better to do this closer to when you generate the report? Imagine that I create the class, then step away from my computer. That'll screw up the results, right?

That's all I've got for now. Sorry to be the eye of Sauron. I think you've got the core pieces here, but making it all make sense is the bigger challenge (particularly when you're generating things like artifacts), and I don't think that's quite there yet. 

I'm going to make a few suggestions on the markdown too, in a separate comment.

mlissner:
OK, and some comments on the comment itself here: https://docs.google.com/document/d/11i9b7GL4Rb77y0Nz7C_AlIDdP17tklwL9Rm0Gjay3yM/edit?usp=sharing

Thank you. This is really cool.

mlissner:
Darn, I knew there was one other thing about the comment. Can you see how hard it is to:

 - [ ] Add a label on the x axis of the chart saying "Citations found" or something like that?

.github/workflows/pr_report.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: |
+          cd benchmark/
+          sh benchmark.sh main ${{ github.head_ref }}
+
+      - name: Embed Timing Chart into Markdown Post
+        run: |
+          echo "![image](https://github.com/freelawproject/eyecite/blob/artifacts/benchmark/pr${{github.event.number}}-time-comparison.png?raw=true)" >> benchmark/outputs/report.md
+
+      - name: Upload and Post Files to PR
+        if: ${{ github.event_name == 'pull_request' }}
```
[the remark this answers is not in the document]
flooie: yep

benchmark/benchmark.md
flooie: yeah lets make this better

benchmark/chart.py
flooie: it shouldn't be ... it was only really needed for testing.

benchmark/chart.py
flooie: this ... is the only thing I couldn't really fix but a comment will be forthcoming

benchmark/chart.py
flooie: yep

flooie:
@mlissner major changes - major slimming down.

benchmark/benchmark.py
```diff
@@ -0,0 +1,209 @@
...
+        benchmark.compare_dataframes()
+
+        # Write Report.MD file
+        benchmark.write_report()
+
+        # Generate time chart
+        benchmark.generate_time_chart()
+    else:
+        # Generate data comparison
+        benchmark.generate_branch_report(branch=args.branch)
```
mlissner, line 209: Looking at this a bit, I wonder if it'd make sense to always require a `--ref` parameter. So, right now the approach is:

 - Check out the current ref
 - Clone main
 - Copy benchmark.py from the current ref to main
 - Run `benchmark.py --branch`
 - Check out the current ref
 - Run `benchmark.py`
 - Run `benchmark.py --chart`

(Or something close to that.)

Would it be impractical to have it change to simply:

 - Run `benchmark.py --ref $SOME_SHORT_HASH`
 - Run `benchmark.py --ref $ANOTHER_SHORT_HASH`
 - Run `benchmark.py --generate-report $SOME_SHORT_HASH $ANOTHER_SHORT_HASH`

And in the `artifacts` branch, instead of storing `main.csv` and `branch.csv` it could store $SHORT_HASH.csv, 

The one tricky thing about this is it moves git stuff into the Python script, but maybe it's reasonable to have that here considering the goal is to compare two hashes?

mlissner (commented):
This is WAY better and simpler. No, it can't easily be run locally, but I think that's a reasonable tradeoff seeing how much nicer this is. I've got a couple more design suggestions for you. Let me know what you think.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,55 @@
+name: Benchmark Pull Request
+on:
+  pull_request:
+
+jobs:
+  benchmark:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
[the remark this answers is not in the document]
mlissner: See this btw: https://github.com/actions/checkout#Checkout-a-different-branch

mlissner:
There are a handful of unresolved comments from my last round too, please take another look at those and see which still make sense when you get a sec.

benchmark/benchmark.py
flooie: im also working to integrate the reporters-db trigger to run the main branch against itself with a reporters-db@HASH install separating the two items.  I think it should be possible to combine them into one test file.

flooie: let me chew on your suggestion.  thats more or less how I had it - or partially had it before - but with branch names... but I realized I could just drop them ... but I do like the idea of storing the hashes, which would let me check if we have that hash output to speed up the entire test by 50% or so

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,55 @@
...
+          virtualenvs-in-project: true
+
+      - name: Run Benchmarking Script
+        run: |
+          curl https://storage.courtlistener.com/bulk-data/eyecite/tests/ten-percent.csv.bz2 --output benchmark/bulk-file.csv.bz2
+          
+          git clone https://github.com/freelawproject/eyecite.git benchmark/eyecite
+          cp benchmark/benchmark.py benchmark/eyecite/ && cd benchmark/eyecite/
+          poetry install --no-dev
+          poetry run pip install matplotlib pandas tabulate
```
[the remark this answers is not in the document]
flooie: cant sneak anything passed you .. part of this is poetry not playing nice with my Mac.  for instance ... and you referenced this before... hypercard doesnt seem to want to play with my apple silicon.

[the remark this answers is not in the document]
flooie: also - this change might need a rebase of a small PR to main adding these dependencies in before we continue - because otherwise checking out the main branch will crash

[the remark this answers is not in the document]
mlissner: Well, this is a particularly interesting PR, so yeah, I'm looking at it closely and thinking a lot about architecture both in this one and in the bigger picture if we build more of these auto-commenters in the future. :)

.github/workflows/benchmark.yml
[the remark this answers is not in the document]
flooie: I think part of this - is we dont have the benchmarking code in any branches.  until then we needed this method to ensure we had the benchmarking code.

PR #122, opened 2022-08-24
Baseline benchmarking

flooie:
A simple small PR to add support to add dev dependencies for benchmarking 

Also includes code to do benchmarking of eyecite.


mlissner (changes requested):
A couple things that jumped out that I made comments for. 

I haven't looked too much at the benchmark.py file yet, but before I do can you please review my earlier comments and integrate them before I do another round?

pyproject.toml
```diff
@@ -49,6 +48,9 @@ hyperscan = ">= 0.1.5"
 exrex = "^0.10.5"
 roman = "^3.3"
 pdoc3 = "^0.10.0"
+tabulate = "^0.8.10"
+matplotlib = "^3.5.3"
+pandas = "^1.4.3"
```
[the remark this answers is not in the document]
flooie: I will try to upgrade my poetry but ... it wouldnt let me create groups today

.github/workflows/tests.yml
```diff
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.8", "3.9", "3.10"]
```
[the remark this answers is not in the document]
flooie: 3.7 was a casualty of pandas.

flooie:
@mlissner I bumped the poetry version to match the latest release - so that groups could be used - and I had to modify the pylint command changing `jsonextended` -> `json`.  setting those two points aside, I think this will let me go back to the original PR and revamp it to simplify it down and answer the remaining comments in the original PR.

pyproject.toml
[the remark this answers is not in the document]
flooie: I added groups - once I bumped the poetry version locally and on CI

.github/workflows/tests.yml
mlissner, line 10: Again, if we're going to do this, we need release notes.

pyproject.toml
```diff
@@ -45,11 +44,18 @@ mypy = "^0.812"
 pylint = "^2.7.2"
 wheel = "^0.35.1"
 pylint-json2html = "^0.3.0"
-hyperscan = ">= 0.1.5"
 exrex = "^0.10.5"
 roman = "^3.3"
 pdoc3 = "^0.10.0"
 
+[tool.poetry.group.x86.dependencies]
+hyperscan = ">= 0.1.5"
```
mlissner, line 52: I don't think we should put this behind a `group`, because if we do, it means older poetry installations won't be able to install it, right?

poetry.lock
```diff
@@ -8,16 +8,17 @@ python-versions = "*"
 
 [[package]]
 name = "astroid"
-version = "2.5.1"
+version = "2.11.7"
```
mlissner, line 11: There are a lot of updates here to deps that aren't part of this. I think you're looking for the `no-update` flag to poetry, IIRC: https://stackoverflow.com/questions/58961497/how-to-update-poetrys-lock-file-without-upgrading-dependencies

benchmark/benchmark.py
```diff
@@ -0,0 +1,224 @@
...
+
+        return: None
+        """
+        row_id = row[0]
+        row = dict(zip(self.fields, row))
+        # Find opinions to test against from the zip file and return
+        # citations found
+        non_empty_rows = [
+            row[field] for field in self.fields if type(row[field]) == str
+        ]
```
mlissner, line 56: Not sure what's happening here. A comment, perhaps?

mlissner (changes requested):
Nice to see that new poetry release candidate is working OK. A couple other comments are below.

Before we merge benchmark.py, we need make sure it's in good shape. A few comments from the other PR that still seem to apply:

> We should be using the cleaning tools before processing the data to strip spaces/html/etc.

> In benchmark.py, there's funky stuff happening with zip(), and self.fields. Why not use a DictReader? Similarly, why create arrays of data that need to be zipped back together instead of creating one array of dicts or objects that have the fields you need? Feels like the data isn't shaped properly for how you are creating it or using it. (I wonder if pandas is the best tool here, but that may be b/c I don't have space for it in my head.)

> In benchmark.py you establish self.now when starting initializing the class. Would it be better to do this closer to when you generate the report? Imagine that I create the class, then step away from my computer. That'll screw up the results, right?

benchmark/benchmark.py
flooie: noted.

pyproject.toml
flooie: I think it depends on whether this is marked as optional or not.  If it isn't set to optional true, it should install automatically, I think.

.github/workflows/tests.yml
flooie: in theory, we should be able to keep 3.7 - I think with groups, but just be aware that benchmarking cant be run on 3.7 but eyecite could.

mlissner: I'm not opposed to dropping 3.7, it's got ten months left until it is no longer supported. Just if we drop it, we need to do it properly is all.

pyproject.toml
mlissner: Interesting. This is worth checking just to be sure. pyproject.toml files have various metadata sections that can be used by lots of tools besides poetry, so I'd be a bit surprised if older poetry versions even read this section, since `tool.poetry.group` didn't exist as a section until recently. But I could be wrong!

flooie: thats probably correct.  I hadn't considered it

mlissner: Only one way to find out. :)

.github/workflows/tests.yml
flooie: I restored 3.7 because ... well its got ten months left.

pyproject.toml
flooie: you are right. about older versions.

flooie:
@mlissner could I get your eyes on this PR, please.

flooie:
I've semi-validated the benchmarking code will work after being merged.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,199 @@
...
+      - name: Setup variables I
+        id: branch1
+        run: |
+          echo ${{ github.event.issue.pull_request }}
+          echo "::set-output name=filepath::benchmark/${{ env.main }}.json"
+          echo "::set-output name=hash::${{ env.main }}"
+
+      - name: Download Testing File
+        run: |
+          curl https://storage.courtlistener.com/bulk-data/eyecite/tests/one-percent.csv.bz2 --output benchmark/bulk-file.csv.bz2
```
mlissner, line 61: We need a comment here or somewhere about how this file was made so we can remake it. I think I shared the details with you on Slack at some point?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,199 @@
...
+      #----------------------------------------------
+      #             Upload to Github PR
+      #----------------------------------------------
+      - name: Pushes test file
+        uses: dmnemec/copy_file_to_another_repo_action@main
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.TOKEN }}
+        with:
+          user_email: 'bill@free.law'
+          user_name: 'flooie'
```
mlissner, line 107: This is a bit weird. Is this required? If so, I guess we have a free law bot Github account, I can resurrect.

.github/workflows/benchmark.yml
flooie: and yes for the record this is what we used.  
```
root@maintenance:/opt/courtlistener# PGPASSWORD=$DB_PASSWORD psql       --command         'set statement_timeout to 0;
           COPY (SELECT id, plain_text, html, html_lawbox, html_columbia, html_anon_2020, xml_harvard FROM search_opinion TABLESAMPLE BERNOULLI (0.1)) TO STDOUT WITH (FORMAT csv, ENCODING utf8, HEADER)'      --host $DB_HOST         --username $DB_USER     --dbname courtlistener  |       bzip2 |         aws s3 cp - s3://com-courtlistener-storage/bulk-data/eyecite/tests/ten-percent.csv.bz2 --acl public-read
root@maintenance:/opt/courtlistener# PGPASSWORD=$DB_PASSWORD psql       --command         'set statement_timeout to 0;
           COPY (SELECT id, plain_text, html, html_lawbox, html_columbia, html_anon_2020, xml_harvard FROM search_opinion TABLESAMPLE BERNOULLI (0.01)) TO STDOUT WITH (FORMAT csv, ENCODING utf8, HEADER)'     --host $DB_HOST         --username $DB_USER     --dbname courtlistener  |       bzip2 |         aws s3 cp - s3://com-courtlistener-storage/bulk-data/eyecite/tests/one-percent.csv.bz2 --acl public-read```

File: .github/workflows/benchmark.yml (hunk 7 above)
Comment (mlissner) [reply .github/workflows/benchmark.yml:107]: OK, you can use this:

```
user_name: "freelawbot"
user_email: "info@free.law"
```

File: .github/workflows/benchmark.yml (hunk 6 above)
Comment (mlissner) [reply .github/workflows/benchmark.yml:61]: Sure, or just put it here as a comment if that's easier. That command looks right to me.

File: .github/workflows/benchmark.yml [hunk 8]
```diff
@@ -0,0 +1,199 @@
...
+      #----------------------------------------------
+      #  -----  install & configure poetry  -----
+      #----------------------------------------------
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+        with:
+          virtualenvs-create: true
+          virtualenvs-in-project: true
+          installer-parallel: true
+          version: 1.2.0rc2
```
mlissner, line 35: We're doing something weird here, so it needs an explanation. That way, in the future, we'll know that we can stop using the `rc` version.

.github/workflows/benchmark.yml
flooie: we also - need to personalize and name the correct secret level to enable some of these features.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,199 @@
...
+      #----------------------------------------------
+      #             Upload to Github PR
+      #----------------------------------------------
+      - name: Pushes test file
+        uses: dmnemec/copy_file_to_another_repo_action@main
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.TOKEN }}
+        with:
+          user_email: 'bill@free.law'
+          user_name: 'flooie'
```
mlissner, line 182: Needs update here too. :)

.github/workflows/benchmark.yml
flooie: Im testing `installer-parallel: true` I think it can be dropped and was/is for matrices

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,199 @@
...
+          source_file: 'results/'
+          destination_repo: 'freelawproject/reporters-db'
+          destination_folder: '${{ github.event.client_payload.pr_number }}'
+          destination_branch: 'artifacts'
+          commit_message: 'Upload to Repo'
```
mlissner, line 187: We should use our usual syntax for this, right? Something like:

```
feat(ci): Add artifacts for PR#XYZ
```

poetry.lock
mlissner: Looks like lots of upgrades to the lock file are back, right?

flooie: I swear- ive never done anything but use --no-update

I promise... but ill go check.

.github/workflows/benchmark.yml
flooie: done

.github/workflows/benchmark.yml
flooie: updated.

.github/workflows/benchmark.yml
flooie: updated

.github/workflows/benchmark.yml
flooie: added

benchmark/benchmark.py
```diff
@@ -0,0 +1,229 @@
...
+    def generate_branch_report(self, branch: str) -> None:
+        """Generate Report
+
+        :param branch: Is a branch from main or not
+        :return: None
+        """
+        zipfile = bz2.BZ2File(self.get_filepath("bulk-file.csv.bz2"))
+        csv_data = csv.DictReader(
+            StringIO(zipfile.read().decode()), delimiter=","
+        )
```
mlissner, line 41: I'm not super familiar with the bz2 library, but I suspect this leaves an open file handle, which, admittedly, doesn't matter since we're running this on some docker image in CI, but still, might I suggest the somewhat tidier:

```suggestion
        with bz2.BZ2File(self.get_filepath("bulk-file.csv.bz2")) as zipfile:
            csv_data = csv.DictReader(zipfile.read().decode(), delimiter=",")
```

I think this should be a bit more performant too, but I'm just guessing.

flooie: this was for legibility

benchmark/benchmark.py
```diff
@@ -0,0 +1,229 @@
...
+        count = 0
+        now = datetime.datetime.now()
+        data = []
+        for row in csv_data:
+            id = row.pop("id")
+            newrow = {k: v for k, v in row.items() if v}
+            found_citations = get_citations(
+                clean_text(
+                    list(newrow.values())[0], ["html", "inline_whitespace"]
```
mlissner, line 50: This is doing some magic to get the first column in the row that has HTML, right? I think it'd be better to be explicit. Try the field you want with early exit on `or` statements, say:

```
text = row['html'] or row['html_lawbox'] or row['plaintext'] or...
```

That spares you from:
 - popping the ID up above
 - filtering out blank columns up above
 - creating newrow
 - casting newrow.values() to a list
 - grabbing the first non-empty value of the list (in a way that's a bit obscure)

benchmark/benchmark.py
```diff
@@ -0,0 +1,229 @@
...
+                )
+            )
+            cites = [cite.token.data for cite in found_citations if cite.token]
```
mlissner, line 53: Can you add a comment here about what this is doing?

benchmark/benchmark.py
```diff
@@ -0,0 +1,229 @@
...
+            count += len(cites)
+            output = {
+                "id": id,
+                "cites": cites,
+                "total": count,
+                "time": (datetime.datetime.now() - now).total_seconds(),
+            }
+            data.append(output)
```
mlissner, line 61: This is so much nicer, thank you!

benchmark/benchmark.py
flooie: its just getting the first non empty opinion.

benchmark/benchmark.py
flooie: 👍

benchmark/benchmark.py
flooie: great.

benchmark/benchmark.py
flooie: this is much nicer

flooie:
not sure why lint - decided to go crazy but... otherwise I think ive addressed your great suggestions

benchmark/benchmark.py
```diff
@@ -0,0 +1,232 @@
...
+        :param branch: Is a branch from main or not
+        :return: None
+        """
+        with bz2.BZ2File(self.get_filepath("bulk-file.csv.bz2")) as zipfile:
+            csv_data = csv.DictReader(zipfile.read().decode(), delimiter=",")
+        count = 0
+        now = datetime.datetime.now()
+        data = []
+        for row in csv_data:
+            id = row.pop("id")
```
mlissner, line 43: I think you can drop this line too. No need to pop it, and `id` is a reserved word in Python we shouldn't be using anyway. This was just to make the `values()[0]` trick work, anyway, right?

mlissner:
OK, review complete, I think. Just drop that `id` call and I think we can call this a wrap. Merge when ready and thank you for refactoring so many times. Sometimes these PRs just take a few extra iterations to get them where they need to be!

benchmark/benchmark.py
flooie: the cluster Id is used in the output to let people review the source material

flooie: but dropped it and used row['id']

mlissner (approved):

benchmark/benchmark.py
mlissner: Yep. :)

flooie:
@mlissner I presume after I fix the lint issue I can merger this.

mlissner:
Yep

PR #123, opened 2022-08-31
fix(benchmark): Bug in zip/bz2 code

flooie:
Small Pr to fix zip reading in benchmarking code


flooie:
The failing test - is expected for this bug.

PR #124, opened 2022-08-31
feat(readme): Update readme to include new CI notes

flooie:
PR to test new access tokens.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -109,7 +109,7 @@ jobs:
       - name: Pushes test file
         uses: dmnemec/copy_file_to_another_repo_action@main
         env:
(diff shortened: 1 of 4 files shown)


flooie:
@mlissner - Im not sure what the issue is but I created a new PA Token under my account and added it to both repositories.   Once I did this both accounts, reporters-db and eyecite began working again.  


1) FREELAWBOT_TOKEN had the wrong credentials scope (unlikely)
2) FREELAWBOT_TOKEN was fine but `freelawbot` didnt have the privileges it needed (possible, but I thought you addressed it) 
3) FREELAWBOT_TOKEN could only be used when triggered by `freelawbot` (possible)

we can rule out three if we approve these two PRs and begin testing all the hanging PRs in reporters-db.

flooie:
Notes. We will need to revisit this for tokens and maybe updates eventually.

PR #125, opened 2022-09-01
fix(workflow): Revert token to freelawbot

flooie:
Revert token for benchmarking workflow


flooie:
All looks good to me.

PR #136, opened 2023-01-13
Fixed parenthesis, added full_span function and test

overmode:
I created a new full_span() function that takes into account the whole citation (including plaintiff, defendant, and post_citation).
We still have an issue when plaintiff name is made of several words, or when the extra attribute contains too much text, but this can be solved independently.
Have a good weekend !

### DIFF
--- a/eyecite/models.py
+++ b/eyecite/models.py
@@ -69,7 +69,9 @@ class CitationBase:
     index: int  # index of _token in the token list
     # span() overrides
     span_start: Optional[int] = None
+    full_span_start: Optional[int] = None
     span_end: Optional[int] = None
+    full_span_end: Optional[int] = None
     groups: dict = field(default_factory=dict)
     metadata: Any = None
 
@@ -144,6 +146,13 @@ class CitationBase:
             if self.span_start is not None
             else self.token.start,
             self.span_end if self.span_end is not None else self.token.end,
+        )
+
+    def full_span(self):
+        """Start and Stop offsets in source text for full citation text (including plaintiff, defendant, post citation, ...)"""
+        return (
+            self.full_span_start if self.full_span_start else self.span()[0],
+            self.full_span_end if self.full_span_end else self.span()[1]
         )
 
 
@@ -262,7 +271,7 @@ class FullLawCitation(FullCitation):
             i for i in (m.publisher, m.month, m.day, m.year) if i
         )
         if publisher_date:
-            parts.append(f" ({publisher_date}")
+            parts.append(f" ({publisher_date})")
         if m.parenthetical:
             parts.append(f" ({m.parenthetical})")
         return "".join(parts)
@@ -358,7 +367,7 @@ class FullCaseCitation(CaseCitation, FullCitation):
             parts.append(f", {m.pin_cite}")
         if m.extra:
             parts.append(m.extra)
-        publisher_date = " ".join(m[i] for i in (m.court, m.year) if i)
+        publisher_date = " ".join(i for i in (m.court, m.year) if i)
         if publisher_date:
             parts.append(f" ({publisher_date}")
         if m.parenthetical:
--- a/eyecite/models.py
+++ b/eyecite/models.py
@@ -69,8 +69,8 @@ class CitationBase:
     index: int  # index of _token in the token list
     # span() overrides
     span_start: Optional[int] = None
-    full_span_start: Optional[int] = None
     span_end: Optional[int] = None
+    full_span_start: Optional[int] = None
     full_span_end: Optional[int] = None
     groups: dict = field(default_factory=dict)
     metadata: Any = None
@@ -149,11 +149,20 @@ class CitationBase:
         )
 
     def full_span(self):
-        """Start and Stop offsets in source text for full citation text (including plaintiff, defendant, post citation, ...)"""
-        return (
-            self.full_span_start if self.full_span_start else self.span()[0],
-            self.full_span_end if self.full_span_end else self.span()[1]
-        )
+        """Span indices that fully cover the citation
+
+        Start and stop offsets in source text for full citation text (including plaintiff, defendant, post citation, ...)
+        Relevant for FullCaseCitation, FullJournalCitation and FullLawCitation.
+        """
+        start = self.full_span_start
+        if start is None:
+            start = self.span()[0]
+
+        end = self.full_span_end
+        if end is None:
+            end = self.span()[1]
+
+        return (start, end)
 
 
 @dataclass(eq=True, unsafe_hash=True, repr=False)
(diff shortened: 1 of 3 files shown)


tests/test_FindTest.py
```diff
@@ -709,3 +709,21 @@ def test_custom_tokenizer(self):
         self.run_test_pairs(
             test_pairs, "Custom tokenizer", tokenizers=[tokenizer]
         )
+
+    def test_citation_fullspan(self):
+        example = 'citation number one is Wilson v. Mar. Overseas Corp., 150 F.3d 1, 6-7 ( 1st Cir. 1998); This is different from Commonwealth v. Bauer, 604 A.2d 1098 (Pa.Super. 1992), my second example'
+        extracted = get_citations(example)
+
+        answers = [(0, (23, 86)), (1, (111, 164))]
```
flooie, line 717: I think this is great, but let's follow the conventions in the other tests. 

I think a better name might be test_FullCaseCitation_fullspan() or something because this doesnt work for other cites right?  

Maybe it does.

flooie: If you could add a docstring with an explanation of the test.

flooie: Also I think it would be more typical for our tests to have a tuples with a test and an output expected

eyecite/models.py
```diff
@@ -146,6 +148,13 @@ def span(self):
             self.span_end if self.span_end is not None else self.token.end,
         )
 
+    def full_span(self):
+        """Start and Stop offsets in source text for full citation text (including plaintiff, defendant, post citation, ...)"""
```
mlissner, line 152: Can you please clean this up. It should be something like:
```
def asdf():
    """One-line description

    More details in a paragraph here.
    """
```

eyecite/models.py
```diff
@@ -69,7 +69,9 @@ class CitationBase:
     index: int  # index of _token in the token list
     # span() overrides
     span_start: Optional[int] = None
+    full_span_start: Optional[int] = None
     span_end: Optional[int] = None
+    full_span_end: Optional[int] = None
```
mlissner, line 74: Just a nitpick, but I think I'd do it this way:

```suggestion
    span_start: Optional[int] = None
    span_end: Optional[int] = None
    full_span_start: Optional[int] = None
    full_span_end: Optional[int] = None
```

mlissner (changes requested):
A couple little things from me. I admit that I really don't understand this code though. I'm not good at this code base. 

I'd feel better if this had another ten test cases. Maybe you can lean on the other tests to make sure they return the correct values?

mlissner:
Also, don't forget to sign our license, please. 

Thanks for this neat feature.

flooie:
@mlissner That being said I think this is actually a great addition that I have wanted before in use.

tests/test_FindTest.py
overmode: It currently works as follows (according to my understanding, please let me know if I'm wrong):
1. Text is tokenized
2. Citation tokens (core citation text, e.g. '150 F.3d 1') are used for instantiating citations
3. For each citation, we fetch metadata using the `add_metadata` function (find.py, l.149)

the `add_metadata` function depends on subclasses :
- `SupraCitation` does nothing
- `IdCitation` does nothing
- `UnknownCitation` does nothing
- `ShortCaseCitation` does nothing
- `CaseCitation` does nothing
- `FullCaseCitation` calls `add_post_citation` and `add_defendant`
- `FullJournalCitation` calls `add_journal_metadata`
-  `FullLawCitation` calls `add_law_metadata`

The `add_x_metadata`, `add_post_citations` functions use regex to find the metadata. I am using the start() and end() functions to know where the regex found interesting parts.
The `add_defendant` function iterates over words and stops after any StopWordToken. Here, I needed to use the number of characters in each word to get an idea of the span.

I modified the `add_defendant`, `add_journal_metadata`, `add_post_citation` and `add_law_metadata` functions.
According to above, the solution should be working for `FullCaseCitation`, `FullJournalCitation` and `FullLawCitation`.
I will try to include examples of these in the tests as well.

So about the name I would like to keep it as it is, what do you think ? Otherwise, it would be nice to avoid a combination of pascal case and snake case.

I will change the docstring and adopt your tuple approach.

overmode:
About the license, should I personally sign it, or should the company I am working at do it ?

overmode:
Fyi this is a list of examples and their spans (full span in red, regular span in blue)
[span_examples.zip](https://github.com/freelawproject/eyecite/files/10432938/span_examples.zip)

mlissner:
> About the license, should I personally sign it, or should the company I am working at do it ?

I think you should sign it, but when you do you're saying that if your work owns the IP, you have the ability to assign that to us, and are doing so.

mlissner:
Sorry, I won't get to this today, but it's on my list for tomorrow. I got bogged down in a major architecture issue, but tomorrow should be better.

overmode:
I signed the contributor license and sent it at legal@free.law, I also pushed a new commit because I realized that stopwords such as 'citing' were included in the span

eyecite/models.py
mlissner: I couldn't find where span_start or span_end are assigned, no. Weird. Could be vestigial. I didn't see span() being tested either, but c.span() is used in the readme, so it seems like these are getting set somehow!

mlissner:
Cool. I cleaned up a couple things, but this looks good to me on the whole. I'm going to merge and hopefully that'll run our benchmark suite, which runs eyecite across a big collection of data (I forget how much). If that runs smoothly, I think we'll be in business. I do wish I knew how `span` was getting set though!

flooie:
@overmode - I want to like your idea of using regexes, but I want to run a few things by you.  In the Pa Super example, our citation string in Courts-db is Pa. Super. Ct.  How would we handle the Ct. citation.  I understand how we do the whitespace but we cant just allow a wild card to handle the ct.

eyecite/models.py
overmode: I was surprised as well, They are probably never set, I think it only works because the `span()` function uses the `start` and `end` attributes of the stored citation's token when `span_start` and `span_end` are not set ([here](https://github.com/freelawproject/eyecite/blob/main/eyecite/models.py#L143-L150))

mlissner: I think you're right. Want to add a fix for this either to remove the attributes or set them?

overmode: Maybe better to set them at initialization to keep backward compatibility

mlissner: Yeah, that's probably better, you're right. Thank you!

mlissner:
@overmode, just a quick update to let you know that 2.5.1 is now released and should have this functionality baked in. Sorry for the delay. Our automated deploy failed and I didn't catch it.

Issue #137, opened 2023-01-19
HyperscanTokenizer erroneously introduces duplicate reporter edition candidates for citations without page numbers


mattdahl:
Just encountered a rather subtle bug in how `HyperscanTokenizer` handles citations without page numbers. (Feature introduced in #116, so I am to blame!) Minimally reproducible version of the problem:

```
# Setup
from eyecite.find import get_citations
from eyecite.tokenizers import HyperscanTokenizer
TOKENIZER = HyperscanTokenizer(cache_dir='.hyperscan')
text = "foo, 574 U. S. ___, bar"

# Returns None
citations = get_citations(text, tokenizer=TOKENIZER)
citations[0].edition_guess

# Returns U.S. (correctly)
citations = get_citations(text)
citations[0].edition_guess
```

The problem occurs because of how `HyperscanTokenizer` deals with citation tokens that are matched by multiple regexes: https://github.com/freelawproject/eyecite/blob/main/eyecite/tokenizers.py#L309. Rather than discarding one match or the other, it attempts to merge them: https://github.com/freelawproject/eyecite/blob/main/eyecite/models.py#L561. However, when the citations' editions are merged, duplicates are not removed. And we only set `edition_guess` if there is only one candidate edition: https://github.com/freelawproject/eyecite/blob/main/eyecite/models.py#L247. So it erroneously records the citation as having two candidate editions, even when those two candidates are the same.

PR to follow...

mattdahl:
After closer review, it appears that the ending comma in the string (e.g., `574 U. S. ___,`) is essential for this bug. I have not fully investigated the root cause, but I think that this may be because the closing comma triggers a match with an additional regex, e.g., the short form regex. (The key to this bug is that the citation token is matched by multiple regexes. Which is not itself a problem, but it currently causes the buggy downstream behavior.)

PR #138, opened 2023-01-19
Issue 137 - Fix Hyperscan duplicate editions bug

mattdahl:
Bug is described in #137. In theory, I think this could affect other tokenizers too, but I only see the problem in `HyperscanTokenizer`. My proposed solution is simple -- just remove duplicate candidate editions after citation tokens are merged. If there is some reason why those duplicates should be retained, I'm missing it.


mlissner:
Seems simple enough, but a number of tests are failing. Any idea what's going on there?

mattdahl:
Not sure what the problem is, they all passed fine on my fork (same commit hash): https://github.com/mattdahl/eyecite/actions/runs/3962150256/. Also not sure why the dependencies (`fast_diff_match_patch`, `exrex`) don't load on only some Python versions.

flooie:
@mattdahl @mlissner - it certainly doesn't appear to be related to your PR.  I re-ran tests for python 3.8 for a branch that passed tests yesterday (I think) to see if it would happen on it now fails.  So somewhere downstream something has changed.

mattdahl:
Can this be reviewed/merged now that #140 has fixed the CI problem? (Though note that now another CI problem will need to be fixed first in #143)

mlissner:
I synced with main and tests are re-running. If they pass, we'll be in business, I think.

mlissner:
Thanks for bumping this Matt! Merged!

PR #143, opened 2023-02-23
Update black

mattdahl:
Black apparently has a new 2023 style that was causing another PR to fail. I implement the fix here, mimicking https://github.com/freelawproject/courtlistener/pull/2510 (though there is no pre-commit in eyecite)


mlissner:
Nice! Thank you.

Issue #135, opened 2023-01-12
Getting full citation span


overmode:
Hi, thank you for the great library !

## Problem description
I am preparing a dataset, in which I would like to _mask_ some citations, e.g. replacing them by "[CITATION]".
I could not find a way to get the _full_ span of the citation. Indeed, only the normalized part is covered by the builtin span() function (see below)  

 ```
 import eyecite         
        
citations = [
    'Commonwealth v. Gibson, 561 A.2d 1240 1242',
    'Commonwealth v. Bauer, 604 A.2d 1098 (Pa.Super. 1992)'
]

for citation in citations :
    print('\n', '='*20)
    extracted_citation = eyecite.get_citations(citation)[0]
    print(extracted_citation)
    
    start_idx = extracted_citation.span()[0]
    end_idx = extracted_citation.span()[1]
    
    before_cit = citation[:start_idx]
    cit_text = citation[start_idx:end_idx]
    after_cit = citation[end_idx:]
    print(f"{before_cit} [BEGIN] {cit_text} [END] {after_cit}")
```

output :

```
====================
FullCaseCitation('561 A.2d 1240', groups={'volume': '561', 'reporter': 'A.2d', 'page': '1240'}, metadata=FullCaseCitation.Metadata(parenthetical=None, pin_cite='1242', year=None, court=None, plaintiff='Commonwealth', defendant='Gibson', extra=None))
Commonwealth v. Gibson,  [BEGIN] 561 A.2d 1240 [END]  1242

 ====================
FullCaseCitation('604 A.2d 1098', groups={'volume': '604', 'reporter': 'A.2d', 'page': '1098'}, metadata=FullCaseCitation.Metadata(parenthetical=None, pin_cite=None, year='1992', court=None, plaintiff='Commonwealth', defendant='Bauer', extra=None))
Commonwealth v. Bauer,  [BEGIN] 604 A.2d 1098 [END]  (Pa.Super. 1992)
```
One can see that the span only partially covers the citation text.
If possible, I would like to avoid using regex for recovering the _full_ span.
Concatenating the lengths of the citation's attributes (plaintiff, defendant, etc.) does not seem to be a viable solution as well, because the second example misses the "Pa. Super" text.

## Desired behavior
It would be nice to have a 'full_span()' function such that, if I use it instead of span() in the above example, I get 
```
====================
[...]

PR #144, opened 2023-02-23
Fix court string matching with whitespace

mattdahl:
As discussed in https://github.com/freelawproject/eyecite/issues/135#issuecomment-1386402484, there is presently a bug where court strings without whitespace are not properly matched. 3b2fe0972ebb86fad1a0beaa9942980398f38cbc implements a failing test for this bug. 94b1e2f16ddbd8005a6c5905a1621850c4adac9c implements a simple fix.

This PR is also related to the changes proposed in https://github.com/freelawproject/eyecite/pull/129, but I think that that proposal has been made obsolete with the removal of all the duplicate citation strings by @flooie (https://github.com/freelawproject/eyecite/issues/135#issuecomment-1404139512). In any case, this PR addresses a different problem re whitespace.

Note that this PR is based off of #143 (needed to update black to make GitHub Actions happy), so that should be merged first.


mlissner (approved):
LGTM. 

My understanding is that in this comment https://github.com/freelawproject/eyecite/issues/135#issuecomment-1404141222, @flooie eliminated duplicate abbreviations from courts DB, so now it's safe to do things like this. 

@flooie, wil you give this a quick review too? The only changes you need to look at are in resolve.py, the rest is updating Black.

Assuming this is good, we need to return to:
 - https://github.com/freelawproject/eyecite/pull/129 (might be obsolete)
 - https://github.com/freelawproject/eyecite/issues/128 (might be fixed)
 - https://github.com/freelawproject/eyecite/issues/135 (don't we already support this, regardless of the code here?)

mlissner:
@flooie looks like this one never got your approval. Mind taking a look, please?

mlissner:
That'd be great, thanks @mattdahl

mattdahl:
Thanks for merging that other PR, @flooie. I just rebased this one as well.

N.B., I previously suggested that #129 had been made obsolete by intervening changes. This is false. More notes over there.

flooie:
Thanks.  I'll take a look soon.

quevon24:
I added a few lines because if you test it with "Commonwealth v. Muniz, 164 A.3d 1189 (Pa. 2017)" as mentioned in the issue, it now returns "njcirctpassaic" instead of "pa", this happens because the first court in the list returned by courts-db that starts with "Pa" is "Passaic Cty. Cir. Ct., N.J."

I added a check to look for an exact match before try the startswith approach, for this specific case this approach works.

mattdahl:
Thanks, I agree that looking for an exact match first is good (cf. https://github.com/freelawproject/eyecite/pull/129#issuecomment-1731981318).

But can we refactor this so we only iterate over `courts` once?

Something like:

```
def get_court_by_paren(paren_string: str) -> Optional[str]:
    """Takes the citation string, usually something like "2d Cir", and maps
    that back to the court code.

    Does not work on SCOTUS, since that court lacks parentheticals, and
    needs to be handled after disambiguation has been completed.
    """
    court_str = strip_punct(paren_string)
    court_str = court_str.replace(" ", "")

    court_code = None
    if court_str:
        for court in courts:
            s = strip_punct(court["citation_string"]).replace(" ", "")
    
            # Check for an exact match first
            if s == court_str:
                return court["id"]
    
            # If no exact match, try to record a startswith match for possible eventual return
            if s.startswith(court_str):
                court_code = court["id"]
    
        return court_code
```

quevon24:
> Thanks, I agree that looking for an exact match first is good (cf. [#129 (comment)](https://github.com/freelawproject/eyecite/pull/129#issuecomment-1731981318)).
> 
> But can we refactor this so we only iterate over `courts` once?
> 
> Something like:
> 
> ```
> def get_court_by_paren(paren_string: str) -> Optional[str]:
>     """Takes the citation string, usually something like "2d Cir", and maps
>     that back to the court code.
> 
>     Does not work on SCOTUS, since that court lacks parentheticals, and
>     needs to be handled after disambiguation has been completed.
>     """
>     court_str = strip_punct(paren_string)
>     court_str = court_str.replace(" ", "")
> 
>     court_code = None
>     if court_str:
>         for court in courts:
>             s = strip_punct(court["citation_string"]).replace(" ", "")
>     
>             # Check for an exact match first
>             if s == court_str:
>                 return court["id"]
>     
>             # If no exact match, try to record a startswith match for possible eventual return
>             if s.startswith(court_str):
>                 court_code = court["id"]
>     
>         return court_code
> ```

great, thanks, it looks more readable, I'll push this change

quevon24:
There is a new PR for all checks to pass correctly: https://github.com/freelawproject/eyecite/pull/188

quevon24:
@flooie this PR is ready, i think that check "Benchmark Pull Request / PR comment (pull_request)" may not be passing because it is a branch of a fork, that is why there may not be permissions to call the workflow that adds the comment with the benchmark

flooie (approved):
Thanks @mattdahl

PR #152, opened 2023-05-23
feat(v): Release new version

mlissner:


mlissner:
Looks like main got a little faster. nice.

mlissner:
Wait, that doesn't make sense. Adding a release tag and some notes shouldn't make main faster. This is just noise! :)

PR #153, opened 2023-05-23
feat(ci): Update benchmark to py3.11

mlissner:


mlissner:
Hey, cool, so in #152, the benchmark took more like 35s. In this one, the test took 30s. Go Python 3.11!

PR #156, opened 2023-07-15
Implement NonopinionCitation deprecation plan (#110)

mattdahl:
In #110, eighteen months ago, we renamed `NonopinionCitation` to `UnknownCitation` and implemented a deprecation warning signaling that support for the old object would end in version 2.5.0. We're now on 2.5.2, so I think it is safe to remove this old code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@ Features:
 - None
 
 Changes:
-- None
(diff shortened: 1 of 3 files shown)


mlissner:
Code looks good, but do you want to add a note to CHANGES.md?

mattdahl:
Good thinking; so updated.

mlissner:
Some CI's are broken, but merging anyway so we can get this in. Thank you.

PR #162, opened 2024-02-03
Update CHANGES

mattdahl:
I just noticed that @flooie did a couple of new releases a few weeks ago (dd9472c9f5184efaaf964df3752e686cb2f19f73 and 2cbe140119559b43b571f65c40a2b7ddb618f4c6). Unfortunately, due to my own negligence, I never updated the CHANGES file to reflect my hashing changes in #155.

This PR memorializes those changes in the CHANGES file (and also adds a line about the changed Python support) by *backdating* them into version `2.5.4`. This might be fine, but it also puts us in the slightly awkward position of having pushed a "patch" version that actually contains some more fundamental changes. Ideally, I think it would have been best to have released version `2.6.0`. (To be clear, this is my own fault  because I never put the changes in the CHANGES.) However, at this point, I'm not sure the best way to proceed? One option is to just merge this PR as-is; the other option would be to also push a minor version update for clarity, though the changes are already baked-in to the previous patch version?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,26 +16,42 @@ Fixes:
 
 ## Current
 
-**2.5.5 - 2024-01-10**
+**2.6.0 - 2024-02-07**
 
-Fixes:
- - 
+Features:
+
+ - Removes Python 3.8 and 3.9 support
+ - Adds Python 3.12 support
 
 Changes:
- - Update dependencies
+
+- The hashing and equality behavior of `CitationBase` objects has changed in subtle ways in order to conform with user intuitions and define previously ill-defined behavior. Most importantly, to compare two `eyecite` objects going forward, simply use the native Python syntax `citation1 == citation2`. You can also take the hash of each object and compare that with identical results: `hash(citation1) == hash(citation2)`. This broad change has several more specific implications:
+  - Citation objects that are created from text with a normalized/corrected reporter are now treated as equal to objects created from text *without* a normalized/corrected reporter (e.g., `1 U.S. 1` versus `1 U. S. 1` are now treated as equal)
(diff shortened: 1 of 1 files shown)


mlissner:
Whoops. I think it's possible to yank a version from pypi, but I don't know how it's done. That might be the best thing, and then we can cut 2.6.0? 

Want to poke around and see how easy it is to yank a version?

mattdahl:
Interesting, this blog post suggest it's pretty easy: https://snarky.ca/what-to-do-when-you-botch-a-release-on-pypi/

The "yanked" release is not _removed_ from pypi, but it is suppressed so `pip` won't resolve to it unless that exact version is specified. How does that sound?

mlissner:
Yeah, that is easy. I'll publish a new version and track the bad one on Monday.

mlissner:
Matt, I was about to yank the bad release, but I got scared that we now rely on it in CL, so maybe before we yank it, we need to release 2.6.0. 

Do you want to tune up this PR for a 2.6.0 release? Then I can merge it, cut the release, get it on Pypi, and then yank 2.5.5 (once 2.6.0 is out)?

mattdahl:
Yes, sounds good, I will do that tomorrow morning!

mattdahl:
Okay, I moved all the previous changes (and the ones newly memorialized here) into a new `2.6.0` section, and I added a "Yanked" note to the previous patches that we will now yank from pypi. Note that there are actually three of these that need yanking: `2.5.3`, `2.5.4`, and `2.5.5`.

Let me know if you need anything else!

mlissner:
OK, this is merged. @erosendo, can you cut version 2.6.0 of eyecite, and then use Bill's new Github Action to upgrade CL's dependencies to use it?

mlissner:
Thank you Matt!

PR #165, opened 2024-02-13
chore(workflow): Updates the pdoc.yml to fix the deploy step

ERosendo:
This PR updates the pdoc.yml file to change the name of the action of the deploy step. It should be `JamesIves/github-pages-deploy-action@v4` according to the [documentation](https://github.com/JamesIves/github-pages-deploy-action?tab=readme-ov-file#getting-started-airplane), but we're using `jamesIves/github-pages-deploy-action@4.5.0 `instead.


mlissner:
Some divergent performance in the chart, but I think it's the underlying hardware, since this shouldn't affect the performance. Onward!

PR #166, opened 2024-02-13
chore(workflow): Downgrade the python version in the pdoc action

ERosendo:
There is a [known issue with pdoc3](https://github.com/pdoc3/pdoc/issues/392#issuecomment-1872475490) that prevents it from functioning correctly with Python 3.12. When attempting to execute the command `pdoc module`, the following error occurs: `ModuleNotFoundError: No module named 'distutils'`. 

As a workaround, temporarily downgrading the python version within the action to `3.11` should allow it to complete successfully.


mlissner:
Always an adventure!

Issue #167, opened 2024-02-19
Hash function raises unexpected json error


cbrew:
Easy to fix. The hash function at the end of utils calls json.dumps, which does not handle datetime values.

Pull request coming.

import eyecite
small = '25 Texas L.Rev. 199'
cit = eyecite.get_citations(small)
hash(cit[0])

mattdahl:
Thank you, this is an alarming bug... It seems that our `test_factories.journal_citation()` function does not mock citations with a realistic `all_editions` field (which should have datetime stuff), so this was not caught in our tests. That seems like a problem in and of itself, but, separately, the `utils.hash_sha256()` function does need to handle non-serializable types... Thanks for the PR.

cbrew:
the change is really simple, add default=str to the json.dumps call in utils.hash_sha256(). 

Making a pull request is harder than I thought, because of lint checks and all that. Working on it slowly.

PR #168, opened 2024-02-19
test and fix for issue #167

cbrew:
Hash function did not handle datetime and failed.

### DIFF
--- a/eyecite/utils.py
+++ b/eyecite/utils.py
@@ -122,7 +122,8 @@ def hash_sha256(dictionary: dict) -> int:
     """
 
     # Convert the dictionary to a JSON string
-    json_str: str = json.dumps(dictionary, sort_keys=True,default=str)
+    # default needed because of dates
+    json_str: str = json.dumps(dictionary, sort_keys=True, default=str)
 
     # Convert the JSON string to bytes
     json_bytes: bytes = json_str.encode("utf-8")
--- a/eyecite/utils.py
+++ b/eyecite/utils.py
@@ -122,7 +122,7 @@ def hash_sha256(dictionary: dict) -> int:
     """
 
     # Convert the dictionary to a JSON string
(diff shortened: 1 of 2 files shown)


mlissner:
Looks like we've got issues with our build unrelated to this PR. @cbrew, do I guess correctly that you'd rather not go down this rabbit hole? If so, I can ask somebody on our staff to do so, or it's possible @mattdahl might be game.

mlissner:
Fair enough. @cbrew, what do you think?

cbrew:
I don’t think Lexis would thank me for spending the time. I was game if it
was going to take only a little time.  So if someone can take it, great.

I’m interested in finding out how far new tech can take us in delimiting
the borders of the citing language, either with or without assumptions
about which passage the citation is pointing to. The chance of having both
ends of the graph edge visible in pile of law goes up fast as its coverage
increases. Having your tool as an experimental tool is a delight.

mlissner:
No worries! One of our devs fixed the testing problem (hopefully). 

Now the only issue is that your code isn't passing the linting rule for a very minor reason. Mind fixing that and then we'll be good to go?

cbrew:
still getting a mystery error about "resource not accessible ny integration'

mlissner:
@ERosendo can you take another look at getting this one over the hump, please? 

Thanks Chris!

ERosendo:
@mlissner I've updated this PR and I've made some minor code changes to avoid unnecessary updates to the poetry.lock file. All tests are passing and the linting check is successful.

I believe the current issue with the **Benchmark Pull Request check** is related to token permissions. Workflows triggered by pull requests from **forked repositories** use a `read-only` token and all other secrets are not available to the workflow run. This means any attempts to write, such as creating comments on the pull request using `GITHUB_TOKEN` or other secrets, will fail.

mlissner:
Excellent, thank you, merging!

PR #172, opened 2024-03-19
fix(model): Tweaks the corrected_citation_full method

ERosendo:
This PR adds missing closing parentheses to the `corrected_citation_full` method in `FullCaseCitation` and `FullJournalCitation` classes. Additionally, I've added a test to prevent similar issues in the future.


mlissner:
Nice. Want to do a release, please?

PR #175, opened 2024-04-04
Add compatibility for hyperscan 0.7.7

jcushman:
I found that eyecite doesn't work with the current version of hyperscan. This fixes the two compatibility errors I found. Github tests are locked to hyperscan version 0.2.0 because of poetry.lock, which I didn't mess with, so this isn't tested by CI.


mlissner:
Looks good, thanks. We'll get a version pushed with the updated poetry files. 

Thank you!

PR #181, opened 2024-05-21
fix(hyperscan): Handles the InvalidError exception during db loading.

ERosendo:
This PR introduces logic to handle the `InvalidError` exception during loading the cached hyperscan db and fixes [#4053](https://github.com/freelawproject/courtlistener/issues/4053). 

According to the documentation, this error (`InvalidError`) is only returned when the function detects an invalid parameter. Since `loaddb` only takes one parameter(the serialized Hyperscan database), this error indicates a problem with the input ByteString. The current fix addresses this by triggering a full database recompile whenever this exception is encountered.

### DIFF
--- a/eyecite/tokenizers.py
+++ b/eyecite/tokenizers.py
@@ -528,6 +528,12 @@ class HyperscanTokenizer(Tokenizer):
                         )
                     except TypeError:
                         hyperscan_db = hyperscan.loadb(cache_bytes)
+                    except hyperscan.InvalidError:
+                        # Skipping hyperscan_db assignment to force a full
+                        # database recompile as the cached version seems to be
+                        # invalid.
+                        pass
+
                     try:
                         # at some point Scratch became necessary --
                         # https://github.com/darvid/python-hyperscan/issues/50#issuecomment-1386243477
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ Changes:
 - None
 
 Fixes:
-- None
+- Strengthens error handling during the loading of the cached Hyperscan database. This ensures that an invalid cache triggers a rebuild.
 
 
 ## Current


eyecite/tokenizers.py
```diff
@@ -528,6 +528,12 @@ def convert_regex(regex):
                         )
                     except TypeError:
                         hyperscan_db = hyperscan.loadb(cache_bytes)
+                    except hyperscan.InvalidError:
```
mlissner, line 531: Hm, is it possible that some error would be silenced because of this, leading a user to think it's working properly when it's not?

mlissner (commented):
One question. If the answer is no, then I think we can probably merge. If it's "maybe", then let's brainstorm if there's a way to make this a bit better.

eyecite/tokenizers.py
ERosendo: > is it possible that some error would be silenced because of this, leading a user to think it's working properly when it's not?

I don't think so, I checked the [python-hyperscan ](https://github.com/darvid/python-hyperscan) code and I noticed the `loadb` function relies on Hyperscan's `hs_deserialize_database` for internal database loading., Here's a fragment of the code: 

```py
hs_error_t err = hs_deserialize_database(buf, length, &db->hs_db);
HANDLE_HYPERSCAN_ERR(err, NULL);
if (PyErr_Occurred())
  return NULL;
return odb;
```

The rest of the code is here: [PyObject *loadb](https://github.com/darvid/python-hyperscan/blob/0103e2a3c4f17eb19b1164b871b31b1d5699f567/src/hyperscan/hyperscanmodule.c#L1247-L1270)


Checking the `hs_deserialize_database` method, we can see that this function returns the error (HS_INVALID) if the sanity checks on the provided serialized data (bytes) or the output database pointer (db) fail. Rebuilding the database guarantees a reliable and usable object for the code's future operations.

Here's a fragment of the [hyperscan method](https://github.com/intel/hyperscan/blob/bc3b191ab56055e8560c7cdc161c289c4d76e3d2/src/database.c#L243-L290):

```c
hs_error_t HS_CDECL hs_deserialize_database(const char *bytes,
                                            const size_t length,
                                            hs_database_t **db) {
    if (!bytes || !db) {
        return HS_INVALID;
    }

    // Decode and check the header
    // The db_decode_header method could also returns HS_INVALID
    hs_database_t header;
    hs_error_t ret = db_decode_header(&bytes, length, &header);
    if (ret != HS_SUCCESS) {
        return ret;
    }

   ... 

   if (db_check_crc(tempdb) != HS_SUCCESS) {
        hs_database_free(tempdb);
        return HS_INVALID;
    }

    *db = tempdb;
    return HS_SUCCESS;
}
```

While the same exception can occur with other Hyperscan methods (like `scan`), this try-except clause only handles the exception when the loadb function raises the error.

mlissner:
Sounds good, thanks. Merging. We can cut a release if we want or just let it be. This is rare enough, I think we can wait, and I think we have another issue about not recognizing citations that we'll need to pull in soon anyway.

PR #183, opened 2024-08-26
ShortCaseCitation.full_span now includes parentheticals

varun-magesh:
Previously, only `FullCaseCitation` objects included parentheticals and antecedent text in the `full_span` method.

For example, in this `ShortCaseCitation`:

```python
>>> test_sentence = "See Sargent, 75 F.3d at 89 (explaining that blah blah blah)."
>>> eyecite.get_citations(test_sentence)[0].full_span()
(13, 26)
```

However, if the same citation was a `FullCaseCitation`, what is included differs:

```python
>>> test_sentence = "See Sargent v. Columbia Forest Prods., 75 F.3d 86 (2d Cir. 1996) (explaining that blah blah blah)."
>>> eyecite.get_citations(test_sentence)[0].full_span()
(4, 97)
```

I am working in an application that requires all `full_span` definitions to be more like the latter, so I've made a small set of changes to `ShortCaseCitation`, `IdCitation`, and `SupraCitation`.

No existing tests are broken by the changes, but I have not written new tests for the behavior.

If this is desired behavior; that is, if it would be helpful to harmonize these definitions by always including the parenthetical and antecedent, I can write some tests and clean this up a bit into a PR. Let me know if that would be helpful.

### DIFF
--- a/eyecite/find.py
+++ b/eyecite/find.py
@@ -175,7 +175,7 @@ def _extract_shortform_citation(
 
     # Get pin_cite
     cite_token = cast(CitationToken, words[index])
-    pin_cite, span_end, parenthetical = extract_pin_cite(
+    pin_cite, span_end, full_span_end, parenthetical = extract_pin_cite(
         words, index, prefix=cite_token.groups["page"]
     )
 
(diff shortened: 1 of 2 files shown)


mlissner:
Makes sense to me. Thank you!

mattdahl:
I agree, this behavior should definitely be the same for full vs. short/supra/id citations.

PR #188, opened 2024-12-11
fix(benchmark): Add permissions to benchmark workflow

quevon24:


quevon24:
I think we don't need this PR anymore. The workflow benchmark has already been working normally.

Issue #196, opened 2025-01-30
Try to include style HTML tags in citation span


grossir:
See parent issue for description

PR #197, opened 2025-01-30
fix(annotate_citations): try to include HTML style tags if not balanced

grossir:
Some annotations, specially for ReferenceCitations, are discarded in HTML sources because some style tags (mostly i or em) are not balanced. This PR tries to include the style tags in the citation span

- Adds tests for `utils.maybe_balance_style_tags`

Solves #196

### DIFF
--- a/eyecite/annotate.py
+++ b/eyecite/annotate.py
@@ -1,11 +1,18 @@
 from bisect import bisect_left, bisect_right
 from difflib import SequenceMatcher
 from functools import partial
+from logging import getLogger
 from typing import Any, Callable, Iterable, Optional, Tuple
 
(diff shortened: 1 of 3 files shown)


grossir:
Unrelated to this specific PR, but, how to run the lint jobs locally?
In other repos we just do `pre-commit run --all-files`
Here I ran from the main directory
```
black .
isort .
```
Shouldn't we have a `.pre-commit` config file?

grossir:
@flooie  this is ready for review in case you want to review it or assigned it to someone else

flooie (approved):
This is nice to have fixed. thanks @grossir

Issue #198, opened 2025-01-30
Find name-only ReferenceCitations


grossir:
Finding ReferenceCitations depend on preceding FullCaseCitations. Currently, we find them using either the defendant or the plaintiff as identified on the FullCaseCitation; and we require a pincite.
https://github.com/freelawproject/eyecite/blob/abfc7f76f5977b620f9143e502db93fdf2ea6065/eyecite/find.py#L157-L167

To find name-only ReferenceCitations, without pincites, we need to account for a different structure. I think this is easiest for HTML / XML opinions, since the intention of referencing is made explicit by `<i>` or `<em>` style tags. Other seemingly obvious structure is the `"In $name,"`. But, see the Halper references that do not conform to `in Halper`. And see the blue highlight for a case where the party names was not supposed to be a reference

![Image](https://github.com/user-attachments/assets/93189a72-df70-4a55-ab56-f9ecd0f487e4)

So, I think we should start looking for this kind of citations on HTML / XML documents only, and using the structured text itself, not a cleaned text. This second round of citation finding may benefit also from using the resolved objects metadata https://github.com/freelawproject/eyecite/issues/199

grossir:
@flooie we talked about this in Monday, I think you were going to implement a PR?

PR #203, opened 2025-02-05
feat(find_reference_citations_from_markup)

grossir:
Solves https://github.com/freelawproject/eyecite/issues/198
Solves problems made evident by first iteration of this PR and described here in #209 

- Implements a function to get name-only ReferenceCitations, taking advantage of style i/em tags on HTML sources
- this new function will be triggered by passing an extra argument to the main function `find.get_citations`
- Refactors ReferenceCitation.is_valid_name to utils.is_valid_name
- adds `regexes.PRE_FULL_CITATION_REGEX` to account for single-name full case citations and for single-name-and-pincite-full-case-citations
- add tests for the new function, to check both that it works as standalone, and that it does not collide with other citation types
- resolved a bug in match_on_tokens where MAX_MATCH_CHARS was used incorrectly
- updated tests that where invalidated, where what was identified as a Reference was actually a part of the FullCaseCitation


quevon24 (approved):
It took me some time to understand the changes and the existing code but I think everything looks good, the code is well commented and structured, I tried the tests and there were no problems.

I think it's ready now

great job @grossir

flooie:
This looks great.  Cant wait to rerun the tests

PR #206, opened 2025-02-06
Improve citation filtering

grossir:


mlissner:
GAINS. 💪🏻

flooie:
@mlissner gains - but at some cost.

eyecite/find.py
```diff
@@ -166,7 +166,7 @@ def is_valid_name(name: str) -> bool:
     if not regexes:
         return []
     pin_cite_re = (
-        rf"\b(?:{'|'.join(regexes)})\s+at\s+(?P<pin_cite>\d{{1,5}})\b"
+        rf"\b(?:{'|'.join(regexes)})\s+at(\s¶)?\s+(?P<pin_cite>\d{{1,5}})\b"
```
grossir, line 169: We should add this `"(\s¶)?"` to other citation types that find pincites, such as Id. cites 
image
Examples here https://github.com/freelawproject/courtlistener/issues/4323#issuecomment-2568317428

eyecite/resolve.py
```diff
@@ -91,22 +91,20 @@ def _filter_by_matching_plaintiff_or_defendant_or_resolved_names(
...
+    for key in ReferenceCitation.name_fields:
+        reference_value = getattr(reference_citation.metadata, key)
+        if reference_value:
+            reference_values.append(reference_value)
+    for citation, resource in resolved_full_cites:
+        full_cite_values = list(
+            [value for value in citation.metadata.__dict__.values() if value]
+        )
+        if set(full_cite_values) & set(reference_values):
+            match_count += 1
```
grossir, line 105: match_count is never used?

eyecite/resolve.py
```diff
@@ -91,22 +91,20 @@ def _filter_by_matching_plaintiff_or_defendant_or_resolved_names(
...
-                and reference_value in full_case_value
-            ):
-                matches.append(resource)
-                break
+    match_count = 0
+    reference_values = []
+    for key in ReferenceCitation.name_fields:
+        reference_value = getattr(reference_citation.metadata, key)
+        if reference_value:
+            reference_values.append(reference_value)
```
grossir, line 99: if this is going to be transformed into a set(), then it could just as well start as one

eyecite/resolve.py
```diff
@@ -91,22 +91,20 @@ def _filter_by_matching_plaintiff_or_defendant_or_resolved_names(
...
+    for citation, resource in resolved_full_cites:
+        full_cite_values = list(
+            [value for value in citation.metadata.__dict__.values() if value]
```
grossir, line 102: this can also be a set, also the `list()` call is unnecessary
`{value for value in citation.metadata.__dict__.values() if value}`

eyecite/helpers.py
```diff
@@ -329,31 +349,30 @@ def filter_citations(citations: List[CitationBase]) -> List[CitationBase]:
...
+                filtered_citations.pop(-1)
+                filtered_citations.append(citation)
+                continue
+            if is_overlapping and isinstance(citation, ReferenceCitation):
+                # Skip overlapping reference citations
                 continue
-        filtered_citations.append(citation)
+            filtered_citations.append(citation)
+        else:
+            filtered_citations.append(citation)
```
grossir, line 375: This could be simplified; we are using a whole indentation level just for the first pass of the for loop

tests/test_FindTest.py
```diff
@@ -730,6 +730,36 @@ def test_date_in_editions(self):
...
+                reporter_found="U.S.",
+                short=False,
+                span_start=26,
+                span_end=38,
+                full_span_start=8,
+                full_span_end=49,
+                metadata={"plaintiff": "Conley", "defendant": "Gibson"},
+            ),
+            reference_citation("Conley", span_start=8, span_end=14),
+            reference_citation("Conley", span_start=18, span_end=24),
```
grossir, line 756: The string here should be "Gibson"

I think it may be better to reproduce the plain example string, and the run `get_citations(test_string)` and test the filtering on top of that

eyecite/helpers.py
```diff
@@ -329,31 +349,30 @@ def filter_citations(citations: List[CitationBase]) -> List[CitationBase]:
...
+            is_overlapping = overlapping_citations(
+                citation.full_span(), last_citation.full_span()
+            )
+            if is_overlapping and isinstance(last_citation, ReferenceCitation):
+                # Remove the overlapping reference citation
+                filtered_citations.pop(-1)
+                filtered_citations.append(citation)
+                continue
+            if is_overlapping and isinstance(citation, ReferenceCitation):
+                # Skip overlapping reference citations
```
grossir, line 371: Refactoring this code shows something strange. If it is overlapping in a full span, but it's not a Reference, we just add it? I think we should drop a `logger.debug` to see what may happen here

```python
if is_overlapping:
     if isinstance(last_citation, ReferenceCitation):
                filtered_citations.pop(-1)
                filtered_citations.append(citation)
                continue
     if isinstance(citation, ReferenceCitation):
                 continue
                 
 filtered_citations.append(citation)
          ...
```

tests/test_FindTest.py
```diff
@@ -882,14 +912,18 @@ def test_reference_extraction(self):
...
-            self.assertEqual(
-                len(final_citations), 2, "There should only be 2 citations"
-            )
-            self.assertEqual(
-                len(references),
-                1,
-                "Only a reference citation should had been picked up",
-            )
+            found_cite = citations[0]
+            if isinstance(found_cite, FullCitation):
```
grossir, line 916: the tests work even when removing this `if isinstance(found_cite, FullCitation):`
Leaving it is just confusing

grossir (commented):
Would have requested changes before you merged this, but it's already done. Probably notes for a refactor

eyecite/find.py
flooie: I think they are covered by the extensive pincite regex already generated for those?  I just didnt start them in the first version of reference citation extraction .

tests/test_FindTest.py
flooie: I dont think we can reproduce it because the bug has been removed - but I wanted to show that if something like this does occur we can fix it.

tests/test_FindTest.py
flooie: yes - but linting requires it

Issue #213, opened 2025-02-14
Bug in Regex generator in Eyecite for variations


flooie:
Bug: Short Case Citation Regex Generator Fails for Many Variations

## Description

Unfortunately, I discovered an old bug in the short case citation regex generator.

```
def short_cite_re(regex):
    """Convert a full citation regex into a short citation regex.
    Currently this just means we turn '(?P<reporter>...),? (?P<page>...'
    to '(?P<reporter>...),? at (?P<page>...'"""

```
Using regex to generate short cite regex was failing for hundreds of variations, causing them to be skipped.

I noticed this when reviewing citation gains and losses that included complex reporter groups with nested parentheses.  This is not uncommon, where a reporter would do something like  `"So (2nd)": "So. 2d",`

Unfortunately, this means that all variations for So. 2d are skipped and only the standard version was used.  Here I noticed So.2d (no whitespace) was not being found when it was a short case citation ... like 8 So.2d at 123, but 8 So.3d at 123 and 8 So. 2d 123 all worked. 

I'm not sure how common this is or what we will see but a simple `replace` should be a good solution.

mlissner:
In other words, if there's a parenthesis in any of the variations for a reporter, it messes up the regex and we fail to match any variations?

flooie:
It would fail any short cite regex variation for an edition, not for the full reporter.

So in the case I stumbled onto, So, So.2d and So.3d - are all part of the southern reporter.  Only So.2d has variations with parentheses in them. so when the variation regex was built it failed and is /was not added to the tokenizer. 

This only affects short cite regex, but as you can see from the small sample run it affects some big reporters.

```
<!DOCTYPE html>
1917661 |   | 581 So.2d at 1030
-- | -- | --
1917661 |   | 589 So.2d at 519
1917661 |   | 669 So.2d at 464
755368 |   | 178 USPQ at 163
1783747 |   | 182 So.2d at 705
1783747 |   | 346 So.2d at 364
1783747 |   | 123 So.2d at 427
1783747 |   | 291 So.2d at 714
1783747 |   | 511 So.2d at 883
1137818 |   | 312 So.2d at 502
1814863 |   | 260 So.2d at 632
1814863 |   | 316 So.2d at 734
2357843 |   | 127 So.2d at 421
280035 |   | 362 F. 2d at 546
280035 |   | 314 F. 2d at 601
1964781 |   | 806 F. 2d at 1160
1964781 |   | 858 P. 2d at 580
[...]

PR #214, opened 2025-02-14
Short cite regex bug

flooie:
Fix for #213 

Eyecite had a bug that could not generate short cite regex patterns for non standard regex variations.  

Anything with content between reporter and page,
Anything with parenthesis in the reporter all caused failures.  

This affected possibly 1537 variants.

### DIFF
--- a/eyecite/regexes.py
+++ b/eyecite/regexes.py
@@ -21,22 +21,7 @@ def short_cite_re(regex):
     """Convert a full citation regex into a short citation regex.
     Currently this just means we turn '(?P<reporter>...),? (?P<page>...'
     to '(?P<reporter>...),? at (?P<page>...'"""
-    return re.sub(
-        r"""
-            # reporter group:
-            (
-                \(\?P<reporter>[^)]+\)
-            )
-            (?:,\?)?\  # comma and space
-            # page group:
-            (
-                \(\?P<page>
-            )
-        """,
-        r"\1,? at \2",
-        regex,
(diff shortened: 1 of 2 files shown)


flooie:
Reminder that these two columns are reversed.  But this is great and should fix many many missing variants.

grossir (commented):
The gains looks great and it solves a silent problem; 

I just found the naming confusing, which hasn't got to do with this PR but with the code base itself

As I understood it, the ShortCaseCitation was different from the FullCaseCitation `<volume> <reporter> <page>` in that it didn't have a page, but a pincite. The pincite most of the times is actually a page; but not the start page that defines the citation

However, we put call the pincite part a page
https://github.com/freelawproject/eyecite/blob/bb9ca00f64c5aa47d0eb85a16e38bc03a6bf0b61/eyecite/regexes.py#L20-L39

And then we will look for another pincite looking part
https://github.com/freelawproject/eyecite/blob/bb9ca00f64c5aa47d0eb85a16e38bc03a6bf0b61/eyecite/regexes.py#L275-L281

But then we pick the page and name it a pincite in `find.extract_short_form_citation`
https://github.com/freelawproject/eyecite/blob/bb9ca00f64c5aa47d0eb85a16e38bc03a6bf0b61/eyecite/find.py#L257-L262

So, perhaps it's a naming problem in `POST_SHORT_CITATION_REGEX`, which is not used for ShortCaseCitation

PR #219, opened 2025-02-18
fix(helpers): Improve NY parsing

flooie:
- Cleanup plaintiff/defendant names with parentheses/whitespace
- Update regexes to allow for square brackets common in NY and California


grossir (approved):
This looks good; I think the plaintiff / defendant basic cleaning will benefit other courts too

PR #224, opened 2025-02-19
feat(pin_cite): Add pin_cite span helper

flooie:
Add a span pincite helper for full case citations

This should make it easier to annotate pincites

### DIFF
--- a/eyecite/helpers.py
+++ b/eyecite/helpers.py
@@ -103,6 +103,9 @@ def add_post_citation(citation: CaseCitation, words: Tokens) -> None:
 
     citation.full_span_end = citation.span()[1] + m.end()
     citation.metadata.pin_cite = clean_pin_cite(m["pin_cite"]) or None
+    if m["pin_cite"]:
+        citation.pin_cite_end=citation.span()[1] + len(m["pin_cite"])
+
     citation.metadata.extra = (m["extra"] or "").strip() or None
     citation.metadata.parenthetical = process_parenthetical(m["parenthetical"])
 
--- a/eyecite/models.py
+++ b/eyecite/models.py
@@ -73,6 +73,7 @@ class CitationBase:
     span_end: Optional[int] = None
     full_span_start: Optional[int] = None
     full_span_end: Optional[int] = None
+    pin_cite_end: Optional[int] = None
     groups: dict = field(default_factory=dict)
     metadata: Any = None
 
@@ -180,6 +181,17 @@ class CitationBase:
                 else self.token.start
             ),
             self.span_end if self.span_end is not None else self.token.end,
+        )
+
+    def span_with_pincite(self):
+        """Start and stop offsets in source text for pincites."""
+        return (
+            (
+                self.span_start
+                if self.span_start is not None
+                else self.token.start
+            ),
+            self.pin_cite_end if self.pin_cite_end is not None else self.token.end,
         )
 
     def full_span(self) -> Tuple[int, int]:


eyecite/models.py
```diff
@@ -73,6 +73,7 @@ class CitationBase:
     span_end: Optional[int] = None
     full_span_start: Optional[int] = None
     full_span_end: Optional[int] = None
+    pin_cite_start: Optional[int] = None
```
grossir, line 76: I would put this inside `citation.metadata`, just to not clutter the `citation` interface
Also, would change the name to `pincite_span_start`, to keep the naming conventions

tests/test_AnnotateTest.py
```diff
@@ -240,3 +240,49 @@ def test_long_diff(self):
...
+            # capture supra pin cite span in html
+            (
+                "Bell Atlantic Corp. </em>v. <em>Twombly, </em>550 U. S. 544 (2007), which discussed... apellate court’s core competency <em>Twombly, </em>550 U. S., at 557. Evaluating... In <em>Twombly</em>, supra, at 553-554, the ",
+                ["550 U. S. 544", "550 U. S., at 557", "supra, at 553-554"],
+            ),
+            # can we wrap the full pincite that appears before
+            # https://www.courtlistener.com/opinion/8524158/in-re-cahill/
+            (
+                "principal residence.” Nobelman at 332, 113 S.Ct. 2106",
+                ["Nobelman at 332, 113 S.Ct. 2106"],
```
grossir, line 275: in this case, it's the same as the full_span when it shouldn't; check upper comments

eyecite/models.py
```diff
@@ -184,15 +185,23 @@ def span(self):
...
+        start = min(
+            filter(
+                None, [self.pin_cite_start, self.span_start, self.token.start]
             ),
-            self.pin_cite_end if self.pin_cite_end is not None else self.token.end,
+            default=self.token.start,
+        )
+        end = max(
+            filter(None, [self.pin_cite_end, self.token.end, self.span_end]),
+            default=None,
```
grossir, line 202: shouldn't this default to the span end or token end?; potentially dangerous to return None, may return the whole string
```python
In [2]: "this is a string"[1:None]
Out[2]: 'his is a string'
```

eyecite/models.py
```diff
@@ -184,15 +185,23 @@ def span(self):
         )
 
     def span_with_pincite(self):
```
grossir, line 187: missing type annotation

eyecite/helpers.py
```diff
@@ -185,6 +185,11 @@ def add_pre_citation(citation: FullCaseCitation, words: Tokens) -> None:
     if not m:
         return
 
+    if m["pin_cite"]:
+        # if a pin cite occurs before the citation mark it down
+        start, end = m.span()
```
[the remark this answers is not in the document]
flooie: yes - I realize that - but made the calculated choice because of hte uniqueness of it and how awkward it would look otherwise.

eyecite/models.py
flooie: sure thanks

tests/test_AnnotateTest.py
flooie: yes, this was a conscious choice here because of the edge-case-y-ness of this example

eyecite/models.py
flooie: split the different pin_cite_span_start

eyecite/helpers.py
```diff
@@ -185,6 +185,11 @@ def add_pre_citation(citation: FullCaseCitation, words: Tokens) -> None:
...
+        citation.pin_cite_start = citation.span()[0] - (end - start)
```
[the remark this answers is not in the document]
flooie: sure - okay,

grossir (commented):
Looks good; you have to solve the lint and it's ready to go

flooie:
@grossir back to you - I moved the pincite span to metadata and updated tests accordingly.

PR #235, opened 2025-03-04
Add fix for hyperscan tokenizer

flooie:
Added an `if not m: continue` check to `extract_tokens()` 
to prevent processing invalid matches that fail when 
re-running the regex on extracted text. 

Previously, Hyperscan detected matches based on byte 
offsets, but some of these did not align properly 
when converted to Unicode string offsets. This caused
 `.match(text[start:end])` to return `None`, 
 potentially leading to errors when calling 
 `get_token(m, offset=start)`. 

Now, we explicitly skip such cases to ensure 
only valid tokens are processed.

Also - put some rationale constraints on the section regex
 and punctuation regexes to avoid matching indefinitely on gibberish.

### DIFF
--- a/eyecite/clean.py
+++ b/eyecite/clean.py
@@ -78,7 +78,9 @@ def all_whitespace(text: str) -> str:
     Returns:
         Text with collapsed whitespace characters.
     """
-    WHITESPACE_REGEX = r"[ \t\n\r\f\v\u00A0\u2002\u2003\u2009\u200B\u202F\u205F]+"
+    WHITESPACE_REGEX = (
+        r"[ \t\n\r\f\v\u00A0\u2002\u2003\u2009\u200B\u202F\u205F]+"
+    )
     return re.sub(WHITESPACE_REGEX, " ", text)
 
 
--- a/eyecite/clean.py
+++ b/eyecite/clean.py
@@ -78,7 +78,8 @@ def all_whitespace(text: str) -> str:
     Returns:
         Text with collapsed whitespace characters.
     """
-    return re.sub(r"\s+", " ", text)
+    WHITESPACE_REGEX = r"[ \t\n\r\f\v\u00A0\u2002\u2003\u2009\u200B\u202F\u205F]+"
+    return re.sub(WHITESPACE_REGEX, " ", text)
 
 
 def underscores(text: str) -> str:
--- a/eyecite/regexes.py
+++ b/eyecite/regexes.py
@@ -52,7 +52,7 @@ PAGE_NUMBER_REGEX = rf"(?:\d+|{ROMAN_NUMERAL_REGEX}|_+)"
 
 # Regex to match punctuation around volume numbers and stopwords.
 # This could potentially be more precise.
-PUNCTUATION_REGEX = r"[^\sa-zA-Z0-9]*"
+PUNCTUATION_REGEX = r"[^\sa-zA-Z0-9]{,3}"
 
 # Regex for IdToken
 ID_REGEX = space_boundaries_re(r"id\.,?|ibid\.")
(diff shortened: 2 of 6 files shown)


mlissner:
Feels like the kind of PR that should really have a test?

eyecite/clean.py
```diff
@@ -78,7 +78,10 @@ def all_whitespace(text: str) -> str:
     Returns:
         Text with collapsed whitespace characters.
     """
-    return re.sub(r"\s+", " ", text)
+    WHITESPACE_REGEX = (
+        r"[ \t\n\r\f\v\u00A0\u2002\u2003\u2009\u200B\u202F\u205F]+"
```
grossir, line 82: the only character in the list that is not included in `r"\s"` is `\u200b`, so I would suggest doing 

`WHITESPACE_REGEX = "[u200b\s]+"`

```
In [18]: [i for i in list("\t\n\r\f\v\u00A0\u2002\u2003\u2009\u200B\u202F\u205F") if not i.isspace()]
Out[18]: ['\u200b']
```

or
```
In [26]: re.sub(r"\s+", "", "\t\n\r\f\v\u00A0\u2002\u2003\u2009\u200B\u202F\u205F")
Out[26]: '\u200b'
```

This 
- to be more clear; 
- because maybe "\s" contains more characters than what you are listing?

eyecite/tokenizers.py
```diff
@@ -466,6 +466,9 @@ def on_match(index, start, end, flags, context):
                 start = byte_to_str_offset[start]
                 end = byte_to_str_offset[end]
                 m = extractor.compiled_regex.match(text[start:end])
+                if not m:
+                    # skip if re-run regex fails to detect match
```
grossir, line 470: This is not expected to be common, right?
Why don't we put a `logger.error` here, so we can analyze what's going on in the edge cases

tests/test_FindTest.py
```diff
@@ -630,6 +630,12 @@ def test_find_citations(self):
                               metadata={'plaintiff': 'Commonwealth', 'defendant': 'Muniz',
                                         'court': 'pa'})]),
             ('Foo v. Bar,  1 F.Supp. 1 (SC 1967)', [case_citation(volume='1', reporter='F.Supp.', year=1967, page='1', metadata={'plaintiff': 'Foo', 'defendant': 'Bar', 'court': 'sc'})]),
+            ('Shady Grove Farms \xa0v Goldsmith Seeds 1 U.S. 1 (1981)', [
```
grossir, line 633: this is an example for the whitespace change; however; this was already covered by the simple r"\s" regex as pointed in the other comment. I think you should try to cover the error cases we have seen. For [example](https://github.com/freelawproject/eyecite/issues/234#issuecomment-2691102991) would still be broken

```python
In [32]: clean_text(' \x08*\x07\x07\u038bþİ\u038b\u202cڋ\u202a-\x14V\u202c\u202c', ["all_whitespace"])
Out[32]: ' \x08*\x07\x07\u038bþİ\u038b\u202cڋ\u202a-\x14V\u202c\u202c'
```

eyecite/regexes.py
```diff
@@ -52,7 +52,7 @@ def short_cite_re(regex):
 
 # Regex to match punctuation around volume numbers and stopwords.
 # This could potentially be more precise.
-PUNCTUATION_REGEX = r"[^\sa-zA-Z0-9]*"
+PUNCTUATION_REGEX = r"[^\sa-zA-Z0-9]{,3}"
```
grossir, line 55: why make these changes? do you have examples or comments on why?

eyecite/regexes.py
```diff
@@ -79,7 +79,7 @@ def short_cite_re(regex):
 )
 
 # Regex for SectionToken
-SECTION_REGEX = r"(\S*§\S*)"
+SECTION_REGEX = space_boundaries_re(r"([\w\.\,\-]*§[\w\.\,\-]*)")
```
grossir, line 82: escape `\` not needed inside `[]`

grossir: From the benchmark, it seems we are losing these?
- `§15.50(a)`
- `[§]52-249a`

PR #240, opened 2025-03-19
Simplify get citation call for markup

flooie:
Add Document object for better citation parsing

With the introduction of markup parsing,
handling multiple parameters became unwieldy.
This PR creates a new `Document` object that encapsulates:

- Plain and markup text
- Span updates for text mapping
- Tokenized words and extracted citation tokens

This should simplify function calls and ultimately enable better HTML extraction.
Also simplify the get citations when using markup text only requiring markup text and cleaning steps 

Currently we must pre-process plaintext before parsing markup.

### DIFF
--- a/eyecite/find.py
+++ b/eyecite/find.py
@@ -389,13 +389,9 @@ def find_reference_citations_from_markup(
     Creating the SpanUpdaters for each full citation will be too slow,
     re-use them if possible
 
-    :param markup_text: HTML or XML source
-    :param plain_text: cleaned text
+    :param document: Document object we are parsing
     :param citations: list of citations found over plain text. The full cites
         will be used to access parties names metadata
-    :param plain_to_markup: a SpanUpdater from plain or clean text to
-        marked up text
-    :param markup_to_plain: a SpanUpdater from marked up text to plain text
 
     :return: a list of ReferenceCitations
     """
--- a/eyecite/find.py
+++ b/eyecite/find.py
@@ -95,10 +95,10 @@ def get_citations(
                     pre = cast(FullCaseCitation, citations[-1])  # type: ignore
                     citation.is_parallel_citation(pre)
 
-            # Check for reference citations that follow a full citation
-            # Using the plaintiff or defendant
-            references = extract_reference_citations(citation, document)
-            citations.extend(references)
+                # Check for reference citations that follow a full citation
+                # Using the plaintiff or defendant
+                references = extract_reference_citations(citation, document)
+                citations.extend(references)
 
         # CASE 2: Token is an "Id." or "Ibid." reference.
         # In this case, the citation should simply be to the item cited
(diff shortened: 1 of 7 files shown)


eyecite/find.py
```diff
@@ -392,11 +374,8 @@ def _extract_id_citation(
 
 
 def find_reference_citations_from_markup(
-    markup_text: str,
-    plain_text: str,
+    document: Document,
```
grossir, line 377: need to update docstring to account for signature changes

eyecite/find.py
```diff
@@ -84,36 +84,30 @@ def get_citations(
...
-                    citation,
-                    plain_text,
-                    markup_text,
-                    plain_to_markup,
-                    markup_to_plain,
-                )
-                citations.extend(references)
+            # Check for reference citations that follow a full citation
+            # Using the plaintiff or defendant
+            references = extract_reference_citations(citation, document)
```
grossir, line 95: Shouldn't this happen only in the case it's a FullCaseCitation? Should be moved an indentation level inside

grossir (commented):
This looks great; just a couple small comments

eyecite/find.py
flooie: Good question, some short case citations could be useful right?  

Smith at 1234  perhaps ?

eyecite/find.py
flooie: thanks

eyecite/find.py
flooie: I think it could be useful and would be more accurate to identify what is likely to be the reference later

flooie:
Im going to suggest we just skip the benchmark because of the difficulty of changing the function calls.

eyecite/find.py
grossir: You would have to change the code for reference extraction (and the model too, probably)  because it will just return if it is called with something other than a FullCaseCitation

https://github.com/freelawproject/eyecite/blob/ba84b132aaa49cf4bbf6175cf393bdbb3bbbf8cd/eyecite/find.py#L172-L173

grossir:
We should change the benchmark call now, if not done, it will basically deactivate the benchmark for every PR until we actually fix it 
Change isn't too long, it's on these lines
https://github.com/freelawproject/eyecite/blob/ba84b132aaa49cf4bbf6175cf393bdbb3bbbf8cd/benchmark/benchmark.py#L55-L67

```python
            if text:
                # Remove XML encodings from xml_harvard
                markup_text = re.sub(r"^<\?xml.*?\?>", "", text, count=1)
                plain_text = ""
                clean_steps =  ["html", "inline_whitespace"]
            else:
                plain_text = row["plain_text"]
                clean_steps = ["all_whitespace"]
                markup_text = ""

            found_citations = get_citations(
                plain_text,
                markup_text=markup_text, 
                clean_steps =clean_steps
            )
            
```

flooie:
im loving the performance gain-  thanks I was working on the benchmark but looks all set now

eyecite/find.py
flooie: why the model?

mlissner:
How'd y'all squeeze out that performance???

Issue #242, opened 2025-04-06
Court detection fails when parenthetical contains month/day


mattdahl:
There is a problem in detecting the court for a full case citation when the court/year parenthetical contains the specific date on which the opinion was issued.

So, this works:
```
t = "Wallace v. Cellco P'ship, No. CV 14-8052-DSF (AS), 2015 WL 13908106, at *7 (C.D. Cal. 2015)"
get_citations(t)
```

Returns: [FullCaseCitation('2015 WL 13908106', groups={'volume': '2015', 'reporter': 'WL', 'page': '13908106'}, metadata=FullCaseCitation.Metadata(parenthetical=None, pin_cite='at *7', pin_cite_span_start=None, pin_cite_span_end=74, year='2015', **court='cacd'**, plaintiff='Wallace', defendant="Cellco P'ship, No. CV 14-8052-DSF (AS", extra=None, antecedent_guess=None, resolved_case_name_short=None, resolved_case_name=None))]

But this does not:
```
t = "Wallace v. Cellco P'ship, No. CV 14-8052-DSF (AS), 2015 WL 13908106, at *7 (C.D. Cal. Feb. 9, 2015)"
get_citations(t)
```

Returns: [FullCaseCitation('2015 WL 13908106', groups={'volume': '2015', 'reporter': 'WL', 'page': '13908106'}, metadata=FullCaseCitation.Metadata(parenthetical=None, pin_cite='at *7', pin_cite_span_start=None, pin_cite_span_end=74, year='2015', **court=None**, plaintiff='Wallace', defendant="Cellco P'ship, No. CV 14-8052-DSF (AS", extra=None, antecedent_guess=None, resolved_case_name_short=None, resolved_case_name=None))]

PR to follow...

PR #243, opened 2025-04-06
Issue 242 - Fix court detection when parenthetical includes date

mattdahl:
As described in #242, there is currently a problem in detecting the court in a full citation when the parenthetical includes the specific date on which the opinion was issued, e.g., `(C.D. Cal. Feb. 9, 2015)`.

The problem is that we currently treat everything *before* the year as a potential court string, so `C.D. Cal. Feb. 9` in the example above. This causes the court lookup in `helpers.get_court_by_paren()` to fail.

This PR fixes this by modifying the `POST_FULL_CITATION_REGEX` regex to include capturing groups for potential months and days. If these are present, we capture them so the `<court>` group remains pristine (i.e., only `C.D. Cal.`). If they are not present, the behavior should be the same as it currently is.

### DIFF
--- a/eyecite/regexes.py
+++ b/eyecite/regexes.py
@@ -162,10 +162,18 @@ PARENTHETICAL_REGEX = r"""
 
 MONTH_REGEX = r"""
     (?P<month>
-        Jan\.|January|Feb\.|February|Mar\.|March|
-        Apr\.|April|May|June|July|Aug\.|August|
-        Sept\.|September|Oct\.|October|
-        Nov\.|November|Dec\.|December
+        January|Jan\.|
+        February|Feb\.|
+        March|Mar\.|
+        April|Apr\.|
+        May|
+        June|Jun\.|
+        July|Jul\.|
+        August|Aug\.|
+        September|Sept?\.|
+        October|Oct\.|
+        November|Nov\.|
+        December|Dec\.
     )
 """
 
--- a/eyecite/regexes.py
+++ b/eyecite/regexes.py
@@ -276,32 +276,43 @@ SUPRA_ANTECEDENT_REGEX = r"""
 # Post full citation regex:
 # Capture metadata after a full cite. For example given the citation "1 U.S. 1"
 # with the following text:
-#   1 U.S. 1, 4-5, 2 S. Ct. 2, 6-7 (4th Cir. 2012) (overruling foo)
+#   1 U.S. 1, 4-5, 2 S. Ct. 2, 6-7 (4th Cir. Jan. 1, 2012) (overruling foo)
 # we want to capture:
 #   pin_cite = 4-5
 #   extra = 2 S. Ct. 2, 6-7
 #   court = 4th Cir.
+#   month = Jan.
+#   day = 1
 #   year = 2012
 #   parenthetical = overruling foo
 POST_FULL_CITATION_REGEX = rf"""
-    (?:  # handle a full cite with a valid year paren:
-        # content before year paren:
+    (?: # handle a full cite with a valid court+date paren:
+        # content before court+date paren:
         (?:
             # pin cite with comma and extra:
             {PIN_CITE_REGEX}?
             ,?\ ?
             (?P<extra>[^(;]*)
         )
-        # content within year paren:
-        [\(\[](?:
-            # court and year:
-            (?P<court>[^)]+)\ {YEAR_REGEX}|
-            # just year:
-            {YEAR_REGEX}
-        )[\)\]]
+        # content within court+date paren:
+        [\(\[] # opening paren or bracket
+        (?:
+            (?:
+                (?P<court>.*?) # treat anything before date as court
+                (?= # lookahead to stop when we see a month or year
+                    \s+{MONTH_REGEX} |
+                    \s+{YEAR_REGEX}
+                )
+            )?
+            \ ?
+            (?P<month>{MONTH_REGEX})?\ ?   # optional month
+            (?P<day>\d{{1,2}})?\,?\ ?      # optional day and comma
+            (?P<year>{YEAR_REGEX})         # year is required
+        )
+        [\)\]] # closing paren or bracket
         # optional parenthetical comment:
         {PARENTHETICAL_REGEX}
-    |  # handle a pin cite with no valid year paren:
+    |  # handle a pin cite with no valid court+date paren:
         {PIN_CITE_REGEX}
     )
 """
--- a/eyecite/regexes.py
+++ b/eyecite/regexes.py
@@ -162,8 +162,10 @@ PARENTHETICAL_REGEX = r"""
 
 MONTH_REGEX = r"""
     (?P<month>
-        Jan\.|Feb\.|Mar\.|Apr\.|May|June|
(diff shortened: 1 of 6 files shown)


tests/test_CourtsTest.py
```diff
@@ -32,6 +32,10 @@ def test_parenthetical_court_parser(self):
             "State Kennedy, 666 P.2d 1316, 1326 (Or. 1983)": "or",
             "State v. Michael J., 875 A.2d 510, 534-35 (Conn. 2005)": "conn",
             "Commonwealth v. Muniz, 164 A.3d 1189 (Pa. 2017)": "pa",
+            "Commonwealth v. Shaffer, 209 A.3d 957, 969 (Pa. Jan. 1, 2019)": "pa",
+            "Sixty-Eight Liquors, Inc. v. Colvin, 118 S.W.3d 171 (Ky. Aug. 2, 2003)": "ky",
+            "Wisniewski v. Johns-Manville Corp., 812 F.2d 81, 83 (3rd Cir. June 30, 1987).": "ca3",
+            "Wallace v. Cellco P'ship, No. CV 14-8052-DSF (AS), 2015 WL 13908106, at *7 (C.D. Cal. Feb. 9, 2015)": "cacd",
```
flooie, line 38: @mattdahl where did you find this format ?

eyecite/regexes.py
```diff
@@ -106,8 +106,10 @@ def short_cite_re(regex):
 
 MONTH_REGEX = r"""
     (?P<month>
-        Jan\.|Feb\.|Mar\.|Apr\.|May|June|
-        July|Aug\.|Sept\.|Oct\.|Nov\.|Dec\.
+        Jan\.|January|Feb\.|February|Mar\.|March|
+        Apr\.|April|May|June|July|Aug\.|August|
+        Sept\.|September|Oct\.|October|
+        Nov\.|November|Dec\.|December
     )
```
flooie, line 113: I think we can improve this regex slightly.  I see you didnt provide abbreviations for June and July.

```
MONTH_REGEX = r"""
     (?P<month>
      Jan(?:uary)?\.?   |
      Feb(?:ruary)?\.?  |
      Mar(?:ch)?\.?     |
      Apr(?:il)?\.?     |
      May\.?            |
      Jun(?:e)?\.?      |
      Jul(?:y)?\.?      |
      Aug(?:ust)?\.?    |
      Sep(?:t(?:ember)?)?\.? |
      Oct(?:ober)?\.?   |
      Nov(?:ember)?\.?  |
      Dec(?:ember)?\.?
     )
"""
```
I found a number of examples of Jun. being abbreviated. 

`…Corp., 2009 U.S. Dist. LEXIS 52214, at *4 (E.D. Pa. Jun. 18, 2009). Put another way, discovery may not …`
`L 179797 (E.D. Pa. Mar. 25, 1999). Ther`
`WL 363921, at *2 (E.D.Pa. Jul. 1,1996). Because a § 2255`

eyecite/regexes.py
```diff
@@ -219,32 +221,43 @@ def short_cite_re(regex):
...
+                    \s+{MONTH_REGEX} |
+                    \s+{YEAR_REGEX}
+                )
+            )?
+            \ ?
+            (?P<month>{MONTH_REGEX})?\ ?   # optional month
+            (?P<day>\d{{1,2}})?\,?\ ?      # optional day and comma
+            (?P<year>{YEAR_REGEX})         # year is required
+        )
+        [\)\]] # closing paren or bracket
```
flooie, line 257: if we are actually going to identify optional month and optional day, we should store this in the metadata right?  

I suppose if we get it we are actually identifying the date filed ?

flooie (changes requested):
@mattdahl this is great.  Can we make a small tweak to the month regex to be more inclusive of June and July and if we are going to capture it we might as well add it to our citation metadata.

eyecite/regexes.py
mattdahl: Excellent, thanks

tests/test_CourtsTest.py
mattdahl: Line 38 is the only real citation, which I got from a brief. I think this format is very common when the citation is to an unreported trial court order (i.e., just `WL` or `LEXIS`).

The citations on lines 35-37 are synthetic, I just made them up to demonstrate the kinds of things we can now catch, but I'm not sure if anyone would actually include the month/day in citations like those.

eyecite/regexes.py
mattdahl: This is a good point, I will store and expose that info

mattdahl:
(1) I edited the month regexes, but didn't follow your proposed format because it would have captured full month names with a period after them (e.g., `January.`). I think my version has the flexibility you wanted, but let me know if it's missing a case you had in mind.

(2) I exposed the captured month/day information on the `metadata` object. I did not, however, do any normalization of this information or expose it on the base `citation` object itself. (For the year data, we store the original captured string under `citation.metadata.year` AND a normalized int version under `citation.year`. I wasn't sure if we should also put the month/day at the top level like that, but since it's not particularly common I decided to just leave it in the `metadata`.)

(3) I considered editing `corrected_citation_full()` to include the month/day in its output when we have it, but I encountered an issue regarding the court normalization that takes place in that function that made me hesitate to do anything before we decide whether that behavior should itself be changed. See https://github.com/freelawproject/eyecite/issues/135#issuecomment-2819212924.

flooie:
@mattdahl 

That looks great.  
I think a larger PR landed and it had an expanded test that included a month and day 

```
            ("Corp. v. Nature's Farm Prods., No. 99 Civ. 9404 (SHS), 2000 U.S. Dist. LEXIS 12335 (S.D.N.Y. Aug. 25, 2000)",
             [case_citation(
                 volume='2000',
                 reporter='U.S. Dist. LEXIS',
                 page='12335',
                 year=2000,
                 metadata={'plaintiff': "Corp.", 'defendant': "Nature's Farm Prods., No. 99 Civ. 9404 (SHS)", "month": "Aug.", "day": "25", "court": "nysd"})
              ],),
```

this is the fixed version, I think if we add that in we can land this

mattdahl:
Thanks, I just fixed that test in 71a46d4e98e53888f588280a74380b5ce546c66c. Also, v2.7.1 just landed so I rebased again.

flooie:
you have permission to run theses test right?  or am I going crazy?

mattdahl:
I think the tests are fine but the benchmark check has never be able to run properly for me, for some reason.

flooie:
@mattdahl I ran it locally and it looks fine.  Can you do one last rebase and we can get this merged today.

mattdahl:
I think it's already fully rebased against main (last commit on main: c1edf7acb50b3b7710839b378e4828fc1e0c7a69)

flooie:
@mlissner I think I need you to merge this.

mattdahl:
I will rebase again

flooie:
@mattdahl sorry

mlissner (approved):

mlissner:
I haven't looked at this in ages, but I just marked it as approved so that my review wasn't needed anymore. @flooie, your call completely on this one and hopefully I'm out of the way now.

flooie (approved):
Looks good

PR #247, opened 2025-04-24
fix(models.Document): identify source text and save a reference in Citation

grossir:
Useful for integration with CL
https://github.com/freelawproject/courtlistener/issues/5458

Also, add a `xml` in `clean.py`


flooie (approved):
Looks great. Thanks @grossir

PR #250, opened 2025-05-03
Hyperscan encoding bug

flooie:
Bad weird encodings and zero width whitespace
were causing issues with hyperscan
tokenzier

A few small tweaks

Remove zero width whitespace
remove tokens without data
only return matches in extractors

### DIFF
--- a/tests/test_FindTest.py
+++ b/tests/test_FindTest.py
@@ -660,8 +660,7 @@ class FindTest(TestCase):
              ),
             # extract no page citations
             ("\n 576 U.S. ___, 13",
-             [case_citation(volume='576', reporter='U.S.', page=None, metadata={'pin_cite': '13',
-                                      "court": 'scotus'})],
+             [case_citation(volume='576', reporter='U.S.', page=None, metadata={'pin_cite': '13', "court": 'scotus'})],
              {"clean_steps": ["all_whitespace"]},
              ),
             # test handling zero width whitespace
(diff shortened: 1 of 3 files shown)


flooie:
Just to clarify this does not remove zero width whitespace, it treats zero width whitespace like all whitespace in the whitespace cleanup method.

tests/test_FindTest.py
```diff
@@ -658,6 +658,21 @@ def test_find_citations(self):
                 case_citation(
                  year=2009, metadata={'defendant': 'K.F.', "year": "2009"})]
              ),
+            # extract no page citations
+            ("\n 576 U.S. ___, 13",
+             [case_citation(volume='576', reporter='U.S.', page=None, metadata={'pin_cite': '13', "court": 'scotus'})],
+             {"clean_steps": ["all_whitespace"]},
+             ),
```
grossir, line 665: This test case actually breaks the main branch, but not because of whitespace... This is probably a bug in the PlaceholderToken that is handled by checking if the token has data; and addressed properly in another PR

tests/test_FindTest.py
```diff
@@ -658,6 +658,21 @@ def test_find_citations(self):
...
+            # test handling zero width whitespace
+            ("Shady Grove Farms \u200bv Goldsmith Seeds 1 U.S. 1",
+             [case_citation(metadata={'defendant': 'Goldsmith Seeds', "plaintiff": 'Shady Grove Farms'})],
+             {"clean_steps": ["all_whitespace"]},
+             ),
```
grossir, line 670: This test case doesn't break `get_citations` even without the changes in this PR. Why add it?

```
from eyecite.tokenizers import HyperscanTokenizer
HYPERSCAN_TOKENIZER = HyperscanTokenizer(cache_dir=".hyperscan")


get_citations("Shady Grove Farms \u200bv Goldsmith Seeds 1 U.S. 1", clean_steps=["all_whitespace"], tokenizer=HYPERSCAN_TOKENIZER)
```

grossir (commented):
I think this works, but not because of the whitespae addition, but because of the checks on empty tokens, which I think are good

Just need to update the changelog


Some tracking notes:

Related to this closed [PR](https://github.com/freelawproject/eyecite/pull/235), but I can't find the Github issue 

A related Sentry [issue](https://freelawproject.sentry.io/issues/6573913691/?project=5257254&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=1)

tests/test_FindTest.py
flooie: yes I think that is right.  I am happy to remove it if you think it best

tests/test_FindTest.py
flooie: Here im not sure I follow.  The zero width whitespace, interferes with processing the case name from the string.  It effectively keeps the stop word from being recognized and you get this weird bug where it drops the stop word and doesnt recognize it.

```
-                       plaintiff='',
+                       plaintiff='Shady Grove Farms',
?                                  +++++++++++++++++

-                       defendant='Shady Grove Farms Goldsmith Seeds',
?                                  ------------------

+                       defendant='Goldsmith Seeds',
```

PR #251, opened 2025-05-05
fix(references): Add addt'l filter for ref.

flooie:
I noticed that we were generating reference citations with longer case named 
supra citations.  because supra citations are expected to be single word antecedents our 
filtering of reference citations was failing and we were linking twice to opinions where only one should
exist.  This simply checks if reference citations are followed by at, v. or supra and if so assumes 
these are not reference citations and should be ignored.  

See
Screenshot 2025-05-05 at 11 16 57 AM


grossir (commented):
While checking the losses (Gain column...) I noticed that we actually fail to pick up a bunch of supra citations. This is not a bug introduced by this PR, but we should probably track it in another issue for future work

For example, in opinion [1431414](https://www.courtlistener.com/opinion/1431414/lugosi-v-universal-pictures/); we lose the "Grant" reference, but can't pick up the supra. The reason is that the supra's antecedent guess is "Inc.", the short form citation is "367 F.Supp", which is ambiguous in this opinion. This may be solved by improving the antecedent guess 
image

eyecite/find.py
```diff
@@ -450,6 +450,11 @@ def find_reference_citations_from_markup(
             end_in_plain = document.markup_to_plain.update(
                 start_in_markup + match.end(1), bisect_right
             )
+            raw_after = document.plain_text[full_end_in_plain:]
+            if re.match(r"^\s*(at|v\.|supra)\s", raw_after):
+                # filter likely bad reference matches
+                continue
```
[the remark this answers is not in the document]
flooie: yes I saw that - and fixed a variant of this in the other PR let me see if I can tackle this variant

eyecite/regexes.py
```diff
@@ -20,7 +20,7 @@ def short_cite_re(regex):
     """Convert a full citation regex into a short citation regex.
     Currently this just means we turn '(?P<reporter>...),? (?P<page>...'
     to '(?P<reporter>...),? at (?P<page>...'"""
-    return regex.replace("(?P<page>", "at (?P<page>")
+    return regex.replace(r"(?P<page>", r"at( p\.?(age)?)? (?P<page>")
```
grossir, line 23: About this, see my comment about centralizing the pincite regexes on the other PR 
https://github.com/freelawproject/eyecite/pull/252#pullrequestreview-2838343586

eyecite/find.py
```diff
@@ -450,6 +451,15 @@ def find_reference_citations_from_markup(
             end_in_plain = document.markup_to_plain.update(
                 start_in_markup + match.end(1), bisect_right
             )
+            raw_after = document.plain_text[full_end_in_plain:]
+            if re.match(r"^\s*(at|v[.s]|supra)\s", raw_after):
+                # filter likely bad reference matches
+                # when matching references citations in markup it is poosible
```
grossir, line 457: Some typos:
refereces -> reference
poosible -> possible

PR #252, opened 2025-05-05
fix(short_cite_re): Add new short cite format

flooie:
A number of courts use at p. 123 format

See: [link](https://www.courtlistener.com/opinion/1431414/lugosi-v-universal-pictures/) for example.
or: [another](https://www.courtlistener.com/opinion/2570714/comedy-iii-productions-inc-v-gary-saderup-inc/)

`(Schumann v. Loew’s Incorporated, supra, at p. 369.)`
`(174 Cal.App.2d at p. 651.) `
`Nimmer, supra, 19 Law & Contemp. Prob. at p. 216.)`

Format seems to be used in California Supreme Courts


grossir (commented):
As the PR is currently, it works, and I don't see any undesired side effects. We could  merge it but I think we should make an improvement

We should centralize the pincite regexes that are spread for better organization, maintanability and easier reading. We have at least these 3 separate regexes that are used in different scenarios but that are trying to do the same thing.

https://github.com/freelawproject/eyecite/blob/464b65566de0c6a4747de0efdc8e5b5c263240a4/eyecite/regexes.py#L19-L23

https://github.com/freelawproject/eyecite/blob/464b65566de0c6a4747de0efdc8e5b5c263240a4/eyecite/regexes.py#L175-L188

https://github.com/freelawproject/eyecite/blob/464b65566de0c6a4747de0efdc8e5b5c263240a4/eyecite/find.py#L199-L201


With this PR, we will pick up short form citations like the one in the example; but we won't pick up pincited references or other variations (see example [below](https://www.courtlistener.com/opinion/10382846/nazaryan-v-femtometrix/?q=%22+at+p.+%22&type=o&order_by=dateFiled+desc&stat_Published=on&page=4)), even when the `PIN_CITE_TOKEN_REGEX` already considers these; just because they are not centralized

image


Separating the PIN_CITE_TOKEN_REGEX into something like this, and then reusing the `PIN_CITE_TOKEN_PREFIX_REGEX` in `find_pincited_reference_citations` and `short_cite_re` should work

```python
PIN_CITE_TOKEN_PREFIX_REGEX = r"""
    # optional label (longest to shortest):
    (?:
        (?:
            (?:&\ )?note|       # note, & note
            (?:&\ )?nn?\.?|     # n., nn., & nn.
            (?:&\ )?fn?\.?|     # fn., & fn.
            ¶{1,2}|             # ¶
            §{1,2}|             # §
            \*{1,4}|            # *
            pg\.?|              # pg.
            pp?\.?              # p., pp.
        )\ ?  # optional space after label
    )?
"""
PIN_CITE_TOKEN_REGEX = rf"""
    {PIN_CITE_TOKEN_PREFIX_REGEX}
    (?:
        # page:paragraph cite, like 123:24-25 or 123:24-124:25:
        \d+:\d+(?:-\d+(?::\d+)?)?|
        # page range, like 12 or 12-13:
        \d+(?:-\d+)?
    )
"""


```

flooie:
@grossir 

I think your suggestion is a good one, but I don't think we need to incorporate all of it.  At least not in this pr.  

Instead, I moved 
```
 pin_cite_re = ( 
     rf"\b(?:{'|'.join(regexes)})\s+at(\s¶)?\s+(?P<pin_cite>\d{{1,5}})\b" 
 ) 
```
to a function in `regexes.py` and I replaced the small pin cite code with our tried and true pin cite regex code.  

This should allow us to keep all the pincite code in the same place.

flooie:
@grossir lots of good gains!   Thanks for the suggestion.

grossir (approved):
Great nice gains; the losses are actually replaced by a full pincite with page range

My only fear was that this would affect Courtlistener's page annotation, but it doesn't, since over there a number will be extracted through regex from `citation.metadata.pin_cite`
https://github.com/freelawproject/courtlistener/blob/2759803baf8dc8def69fe7f6c4a82cb46d88c80c/cl/citations/annotate_citations.py#L66-L67

PR #259, opened 2025-05-14
fix(helpers): A citation in a parenthetical should not emit the warning "Unknown overlap case"

branliu0:
Previously, a citation like "X v. Y 525 F. App’x 94, 103 n.15 (3d Cir. 2013) (quoting Iqbal, 556 U.S. at 686-87)" would lead to a warning "Unknown overlap case."

However this warning was written with the expectation that overlapping cases would be parallel citations, and that all parallel citations should be FullCaseCitations.

This PR updates the code to handle the case when there is a citation in a parenthetical, in which case the two cases would be overlapping, but do not have to both be FullCaseCitations.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ Changes:
 Fixes:
 - Add fix for reference citation filtering out bad matches
 - Fix for hyperscan bug and whitespace bug
+- A citation in a parenthetical should not emit the warning "Unknown overlap case"
 
 ## Current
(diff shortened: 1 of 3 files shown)


flooie (approved):
Thank you @branliu0 This looks good to me.

flooie:
@branliu0 we will release this later today.

branliu0:
very cool, thank you!

Issue #264, opened 2025-05-15
Incorrect parsing of footnote short cite when short name is also a reporter


branliu0:
In RECAP document [139032416](https://www.courtlistener.com/api/rest/v3/recap-documents/139032416/), there is a footnote `21 Taylor, 281 F.3d at 850 (finding that...)`

Similarly, in RECAP document [139032384](https://www.courtlistener.com/api/rest/v3/recap-documents/139032384/) there is a footnote `28 Phillips, 415 F.3d at 1321 (quoting Vitronics Corp...)`.

Both of these are not parsed as expected:

```python
>>> eyecite.get_citations("21 Taylor, 281 F.3d at 850 (finding that")
[FullCaseCitation('21 Taylor, 281', groups={'volume': '21', 'reporter': 'Taylor', 'page': '281'}, 
metadata=FullCaseCitation.Metadata(parenthetical=None, pin_cite=None, pin_cite_span_start=None, pin_cite_span_end=None, year=None, 
month=None, day=None, court=None, plaintiff=None, defendant=None, extra=None, antecedent_guess=None, 
resolved_case_name_short=None, resolved_case_name=None))]
```

```python
>>> eyecite.get_citations("28 Phillips, 415 F.3d at 1321 (quoting Vitronics Corp.")
[FullCaseCitation('28 Phillips, 415', groups={'volume': '28', 'reporter': 'Phillips', 'page': '415'}, 
metadata=FullCaseCitation.Metadata(parenthetical=None, pin_cite=None, pin_cite_span_start=None, pin_cite_span_end=None, year=None, 
month=None, day=None, court=None, plaintiff=None, defendant=None, extra=None, antecedent_guess=None, 
resolved_case_name_short=None, resolved_case_name=None))]
```

I've never looked at the parsing logic. I suppose one solution might be: if there's a comma after the first assumed reporter name, try looking ahead to see if there's a citation immediately following.

grossir:
this is coming from the line numbers at the margin of the page. 

We could also clean the line numbers as a pre processing step

![Image](https://github.com/user-attachments/assets/8267121e-8c74-49e5-9558-d539d0377655)

![Image](https://github.com/user-attachments/assets/b0e8dbab-e3be-4679-ad22-39b566c0f68d)

branliu0:
Ah! I should have looked at the original document. I just assumed. Thank you for checking!

[...]

PR #267, opened 2025-05-16
fix(tokenizer): Correct parsing of line number + short cite

branliu0:
Fixes https://github.com/freelawproject/eyecite/issues/264

Would appreciate feedback on:
1) naming
2) placement of the helper nested function inside `tokenize`.

Let me know what you think!

### DIFF
--- a/eyecite/tokenizers.py
+++ b/eyecite/tokenizers.py
@@ -388,7 +388,7 @@ class Tokenizer:
         # where the current start offset is greater than the previously
         # returned end offset. Also return text between matches.
         # filter out empty tokens cause by corrupted/complex pdf data
-        citation_tokens = []
+        citation_tokens: list[tuple[int, Token]] = []
         all_tokens: Tokens = []
         tokens = sorted(
             (t for t in self.extract_tokens(text) if t.data is not None),
--- a/eyecite/tokenizers.py
+++ b/eyecite/tokenizers.py
@@ -336,6 +336,49 @@ class Tokenizer:
     def tokenize(self, text: str) -> tuple[Tokens, list[tuple[int, Token]]]:
         """Tokenize text and return list of all tokens, followed by list of
         just non-string tokens along with their positions in the first list."""
+
+        def _is_short_cite_following_line_number(last_token: Token | None, token: Token) -> bool:
+            """
+            When we have a short cite following a line number, it can result in two
+            CitationTokens.
+            e.g., "21 Taylor, 281 F.3d at 830" will result in:
+            1) "21 Taylor, 281"
+            2) "281 F.3d at 830"
+
+            When this occurs, we want to take the latter. However we'll try to be
+            conservative and only take the second token if:
+            - last_token.volume < 50 (line numbers rarely go past 50)
+            - last_token.page == token.volume
+            - the overlap is precisely the last_token's page and the next token's volume
+            - last_token's reporter is followed by a comma, which makes it more likely
+            that we're dealing with a short cite.
+
+            Can we also check whether `token` is a short cite?
+            """
+            if last_token is None:
+                return False
+            if not isinstance(last_token, CitationToken) or not isinstance(token, CitationToken):
+                return False
+            last_token_volume = last_token.groups.get("volume")
+
+            # Is the last token's volume under 50?
+            if not last_token_volume or int(last_token_volume) >= 50:
+                return False
+
+            # Does the last token's page match the next token's volume?
+            last_token_page = last_token.groups["page"]
+            token_volume = token.groups.get("volume")
+            if not token_volume or last_token_page != token_volume:
+                return False
+
+            # Do they precisely overlap on the last token's page/next token's volume?
+            if token.start + len(last_token_page) != last_token.end:
+                return False
+
+            # Is the last token's reporter followed by a comma?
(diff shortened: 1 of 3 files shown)


tests/test_FindTest.py
```diff
@@ -1122,6 +1122,80 @@ def test_nominative_reporter_overlaps(self):
...
+            # Note: Need to remove the editions from the case_citation, because the
+            # parsed_cite object doesn't have any edition attached. When the correct
+            # edition is attached, the correct_reporter() becomes "Tay." and then the
+            # two objects don't match.
+            (
+                # Volume number too large
+                "212 Taylor, 281 F.3d at 830",
+                case_citation(
+                    volume="212", reporter="Taylor", page="281", **no_ed
+                ),
```
flooie, line 1160: Are you suggesting the correct citation here is 212 Taylor.  
```
    "Tay.": [
        {
            "cite_type": "state",
            "editions": {
                "Tay.": {
                    "end": "1802-12-31T00:00:00",
                    "start": "1798-01-01T00:00:00"
                }
            },
            "mlz_jurisdiction": [
                "us:nc;supreme.court"
            ],
            "name": "North Carolina Reports, Taylor",
            "variations": {
                "N.C.(Tay.)": "Tay.",
                "Tay.J.L.": "Tay.",
                "Tay.N.C.": "Tay.",
                "Tayl.N.C.": "Tay.",
                "Taylor": "Tay."
            }
        }
    ],
```
While Taylor is a reporter it is an ancient one and obviously not the correct one.  Reporter volumes routinely reach `999` so im not sure what you mean by reporter volume number too large

eyecite/tokenizers.py
```diff
@@ -336,12 +336,59 @@ class Tokenizer:
...
+            When we have a short cite following a line number, it can result in two
+            CitationTokens.
+            e.g., "21 Taylor, 281 F.3d at 830" will result in:
+            1) "21 Taylor, 281"
+            2) "281 F.3d at 830"
+
+            When this occurs, we want to take the latter. However we'll try to be
+            conservative and only take the second token if:
+            - last_token.volume < 50 (line numbers rarely go past 50)
+            - last_token.page == token.volume
```
flooie, line 353: this is certainly an interesting strategy, I understand now that why you said volume too large.  I think the appropriate way to handle page numbers is not to change how we process citations but improve how we clean the text.  Perhaps a flag for cleaning page line numbers would be more effective approach.  


If I was choosing between overlapping citations 

"21 Taylor, 281"
and 
"281 F.3d at 830" 

I would try to use the one that is common.

flooie:
I took a look at your example line 

`21   Taylor, 281 F.3d at 850 (finding that 28 U.S.C. § 1915(b)(4) acts as a “safety-valve”"`

and it actually produces the correct result if you choose not to strip whitespace.  Honestly that may be the correct approach here.

branliu0:
Thank you for the review, @flooie!

RE: cleaning line numbers instead: I agree, as the line numbers are causing issues with other citations as well. For instance [this document](https://www.courtlistener.com/api/rest/v3/recap-documents/139021810/) ([link to pdf](https://ia803203.us.archive.org/16/items/gov.uscourts.casd.672118/gov.uscourts.casd.672118.17.0.pdf)) has several citations that go across a line break.

I can work on a PR that attempts to clean up these line numbers, but would you be interested in landing this PR here regardless? This code already helps a little bit, and the line number cleaning routine would have to be conservative and therefore may not correctly clean numbers in every single instance. (I'll look into both a text-based cleaning routine and an OCR-cropping approach)

RE: picking between the reporter that is most common -- that could also work, but I feel a bit more wary of that approach, because now the parser is also taking into account "reporter commonality," which feels a bit out of scope for a parser/tokenizer? I mean, it's doable... but I'm not sure if a tokenizer should be that smart.

RE: stripping whitespace -- that's an interesting observation. However, the problem is that not stripping whitespace could result in some other citations going undetected. See, e.g.,

```python
>>> eyecite.get_citations("X v. \n\n  Y, 281 F.3d 850 (finding")
[FullCaseCitation('281 F.3d 850', groups={'volume': '281', 'reporter': 'F.3d', 'page': '850'}, metadata=FullCaseCitation.Metadata(parenthetical=None, pin_cite=None, pin_cite_span_start=None, pin_cite_span_end=None, year=None, month=None, day=None, court=None, plaintiff='X', defendant='Y', extra=None, antecedent_guess=None, resolved_case_name_short=None, resolved_case_name=None))]

>>> eyecite.get_citations("X v. \n\n  Y, 281 F.3d \n 850 (finding")
[]
```

tests/test_FindTest.py
branliu0: I'll update the comment here to say "volume number too large to be a line number, so proceed with taking Taylor as the reporter"

branliu0:
To add to the line number cleaning approach: It'd only work when the full page of text is provided, as I'd try to look for all the line numbers starting from 1. If only an excerpt is provided, I'm not sure we can be confident enough to clean the line numbers. I think it's better to be conservative?

branliu0:
Had a chat with @flooie -- closing this in favor of an alternate approach of cleaning up line numbers instead.

flooie:
Thanks for the chat @branliu0 
cant wait to see more work down the road

Issue #268, opened 2025-05-15
Incorrect Annotation tags in footnotes on opinion


flooie:
In the footnotes of the opinion at https://www.courtlistener.com/opinion/6244094/kubbernus-v-ecal-partners-ltd/, there appears to be a rendering or annotation bug where citation tags are malformed. Specifically, the text shows erroneous <itation> and <ion> tags instead of properly formed <citation> tags.

```
See ion index="148" url="https://cite.case.law/citations/?q=47%20U.S.C.%20%C2%A7%20214">id. art. 581-33(D)(7) (attorney's fees). Appellees brought additional claims against other defendants who are not parties to this appeal. ↵
See itation index="149" url="https://cite.case.law/citations/?q=47%20U.S.C.%20%C2%A7%20214">id. art. 581-33(D)(3), (6), (7). ↵
See itation index="150" url="https://cite.case.law/citations/?q=47%20U.S.C.%20%C2%A7%20214">id. ↵
See itation index="151" url="https://cite.case.law/citations/?q=47%20U.S.C.%20%C2%A7%20214">id. ↵
```

Need to do a deeper look at if this is a one/off or not .  I havent seen this before.

flooie:
#### Scope / Frequency:

Found ~100 occurrences in just the first three courts I checked (texapp, mass, cal).
Affects any HTML that uses id or index (or other attributes starting with “i”) adjacent to visible tokens like id. or if style tags exist for “supra” references.

#### Root cause:
Our HTML-unaware diff runs straight on the raw tag strings. fast match path code splits on every character, so an attribute name beginning with “i” (e.g. id=) introduces an extra “i” chunk that doesn’t appear in the cleaned “plain” text. Offsets diverge and insertions float into the tag.

#### Rejected alternatives:
Stripping out problematic attributes before diff: loses metadata and causes formatting issues for us and sorta is unwelcome for anyone who wants to annotate their own text.
Modifying the diff tool to be aware of tags felt like its own can of worms and probably would be too slow

#### Solution
On Document instantiation, replace every tag (or every tag with any attribute starting with “i”) with a placeholder token (e.g. <XXXXX>) and store it in the plain_to_markup span updater.
Run the HTML-aware diff against those placeholders—plain text vs. placeholder markup will align perfectly.
We preserve our markup text and this is only used to jump back and forth.  It does not affect anything else.

[...]

PR #269, opened 2025-05-20
feat(annotation): Update annotate citations

flooie:
Update annotation citations 

Create a method to remove tags from html when generating annotations
to avoid mapping bug with supra and id citations.

### DIFF
--- a/eyecite/annotate.py
+++ b/eyecite/annotate.py
@@ -10,6 +10,7 @@ import fast_diff_match_patch
 from eyecite.utils import (
     is_balanced_html,
     maybe_balance_style_tags,
+    placeholder_markup,
     wrap_html_tags,
 )
 
@@ -127,7 +128,6 @@ def annotate_citations(
     unbalanced_tags: str = "unchecked",
     use_dmp: bool = True,
     annotator: Optional[Callable[[Any, str, Any], str]] = None,
-    offset_updater: Optional[SpanUpdater] = None,
 ) -> str:
     """Given a list of citations and the text from which they were parsed,
     insert annotations into the text surrounding each citation. This could be
@@ -168,21 +168,16 @@ def annotate_citations(
             useful for customizing the annotation action: If you don't pass
             this function, eyecite will simply concatenate the before_text,
             citation_text, and after_text together for each annotation.
-        offset_updater: If provided, use this SpanUpdater. Citation finding
-            for HTML / XML sources use a SpanUpdater called `plain_to_markup`
-            to find citations, passing it saves this expensive instantiation.
     Returns:
         The annotated text.
     """
     if unbalanced_tags not in ["unchecked", "skip", "wrap"]:
         raise ValueError(f"Unknown option '{unbalanced_tags}")
 
-    # set up offset_updater if we have to move annotations to source_text
-    if offset_updater:
-        plain_text = source_text
-    elif source_text and source_text != plain_text:
-        offset_updater = SpanUpdater(plain_text, source_text, use_dmp=use_dmp)
-        plain_text = source_text
+    markup_with_placeholders = placeholder_markup(source_text)
+    offset_updater = SpanUpdater(
+        plain_text, markup_with_placeholders, use_dmp=use_dmp
+    )
 
     # append text for each annotation to out
     annotations = sorted(annotations)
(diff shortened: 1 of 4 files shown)


eyecite/annotate.py
```diff
@@ -127,7 +128,6 @@ def annotate_citations(
     unbalanced_tags: str = "unchecked",
     use_dmp: bool = True,
     annotator: Optional[Callable[[Any, str, Any], str]] = None,
-    offset_updater: Optional[SpanUpdater] = None,
```
grossir, line 130: not re-using this means a peformance penalty on creating a new span updater every time

eyecite/utils.py
```diff
@@ -315,3 +315,29 @@ def maybe_balance_style_tags(
...
+    This allows diffing or annotation algorithms to maintain correct
+    character offsets by hiding tags behind 'X's of the same length.
+
+    Args:
+        html: The raw HTML string to sanitize.
+
+    Returns:
+        A safe string to annotate
+    """
+    tag_re = re.compile(r"<([\/a-z])[^>]+>")
```
grossir, line 332: this won't capture any single letter tag, like `<a>` or `<p>`, unless they have attributes. So, you are really trying to capture attributes? This should be more clear on the docstring

Also, if attributes are the problem, why not clean up the problematic attributes from xml harvard? and prevent reprocessing all other documents

grossir (commented):
This PR addresses this issue https://github.com/freelawproject/eyecite/issues/268

I think your code works; I am not sure if it will have unintended side effects. Some things to consider
- do you have more examples of where this is happening? so far, in the issue there is a single opinion; seems to be a bug specific of xml sources
- Have you tried solving this by deleting "index" attributes or other problematic sources?
- the creation of a new span updater will cause a performance penalty; we were already saving a few processing steps by re-using the SpanUpdater created in the Document

From your solution and the sample code below, this seems to happen with "index" attributes and "id." citations . I think this is a bug with the SpanUpdater class; maybe it's computing an offset between "i" from the "index" attribute and "i" from "id"

```python
from eyecite.annotate import SpanUpdater
from bisect import bisect_right, bisect_left

origin_text = """
See</em> <em>Bacon Tomsoms, Ltd. v. Chrisjo Energy, Inc.</em> , No. 01-15-00305-CV, <extracted-citation index=\"143\" url=\"https://cite.case.law/citations/?q=2016%20WL%204217254\">2016 WL 4217254</extracted-citation>, at *15 (Tex. App.-Houston [1st Dist.] Aug. 9, 2016, no pet.) (mem. op.). The trial court then awarded less than the full amount of attorney's fees requested by appellees. <em>See <extracted-citation index=\"144\" url=\"https://cite.case.law/citations/?q=2016%20WL%204217254\">id.</extracted-citation></em></p>\n<p id=\"p-238\">We conclude that appellees sufficiently segreg
"""

cites = get_citations(markup_text=origin_text, clean_steps=["html", "all_whitespace"])

# Citations are correctly found in the `plain_text` space
start, end = cites[0].span()
document = cites[0].document
print(document.plain_text[start:end])
Out[68]: '2016 WL 4217254'


offset_updater_plain_to_markup = SpanUpdater(document.plain_text, document.source_text)

# mimic annotate code, it works...
new_start = offset_updater_plain_to_markup.update(start, bisect_right)
new_end = offset_updater_plain_to_markup.update(end, bisect_left)

print(document.source_text[new_start:new_end])
Out[5]: '2016 WL 4217254'


# the found span is correct
start, end = cites[-1].span()
print(document.plain_text[start:end])
# 'id.'

# the updated span is wrong and gives a nonsense update
new_start = offset_updater_plain_to_markup.update(start, bisect_right)
new_end = offset_updater_plain_to_markup.update(end, bisect_left)
print(repr(document.source_text[new_start:new_end]))
# 'index="144" url="https://cite.'

# but, if we invert the bisect we get a more coherent answer (that still won't work)
new_start = offset_updater_plain_to_markup.update(start, bisect_left)
new_end = offset_updater_plain_to_markup.update(end, bisect_right)
print(repr(document.source_text[new_start:new_end]))

# '<extracted-citation index="144" url="https://cite.case.law/citations/?q=2016%20WL%204217254">id.</extracted-citation></em></p>\n<p'
```

eyecite/utils.py
flooie: I dont want to clean up the attributes because in some cases we actually use the attributes, specifically `id` attributes to layout the text.  And yes it is only the tags with attributes that could cause an issue.  

I did think about stripping out all the properties, but It seemed cleaner this way.

eyecite/annotate.py
flooie: I think I have an option to satisfy us both.

PR #275, opened 2025-05-27
docs: Simplify hyperscan installation instructions

adamchainz:
Back when the hyperscan installation docs were written in bb741e0daa5d913ae01571f8aecbee188cf087c7 (2021), hyperscan was a bit difficult to install because it required libhyperscan on the system. But since hyperscan version 0.3.0 (2022), it has statically linked libhyperscan, so it’s not much different from other PyPI packages. This commit simplifies the documentation and drops our CI installation of libhyperscan-dev.

### DIFF
--- a/.github/workflows/pdoc.yml
+++ b/.github/workflows/pdoc.yml
@@ -20,9 +20,6 @@ jobs:
           virtualenvs-create: true
           virtualenvs-in-project: true
 
-      - name: Install hyperscan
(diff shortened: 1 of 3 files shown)


mlissner:
One more thing, sorry @adamchainz. Can you remove it from list.yml too?

adamchainz:
Woops, yes, done in #276.

PR #278, opened 2025-06-15
Refactor Document model to simplify public-facing API

mattdahl:
In #240, @flooie introduced a new `Document` class to better keep track of any markup and cleaning steps applied to the text given to eyecite for parsing. However, the class is currently only used internally, meaning that it is still pretty confusing for the user (in my opinion) to keep track of what parameters they should be passing when.

For example, `get_citations()` currently accepts both a `plain_text` and a `markup_text` parameter, but both are typed optional. So can the user choose which of these to pass, or is the user always supposed to pass `plain_text` no matter what? If `markup_text` exists, should it be passed in addition to `plain_text`, or instead of it? Should `plain_text` still be cleaned by the user in advance? But isn't the `Document` abstraction supposed to keep track of the cleaning steps now? Or is that cleaning stage only cleaning for `markup_text`, and if there's other cleaning to be done, the user should still handle that in advance? And when the user later calls `annotate_citations()`, they're supposed to remember what the original `plain_text` was (along with the `source_text`, an entirely new variable?), and pass some combination of it all over again?

This PR attempts to resolve these confusions by simply exposing the internal `Document` model to the user. The idea is that the user should always instantiate a `Document` himself, and then just pass *that object* to `get_citations()` and `annotate_citations()`. All the pre-processing related parameters (whether the text has markup, any cleaning steps, what diffing algorithm to use, etc.) are dealt with only *once* at the `Document` level and then never needed again (from the user's perspective).

So with this PR, for example, the user would do:

```
from eyecite import get_citations, annotate_citations, Document

document = Document(
    opinion_text,
    has_markup=True,
    clean_steps=["html", "all_whitespace"]
)
citations = get_citations(document)
linked_text = annotate_citations(
     document,
     annotations=[[c.span(), "<a>", "</a>"] for c in citations]
)
```

Instead of (current syntax):

```
from eyecite import get_citations, annotate_citations, clean_text

markup_text = opinion_text
plain_text = clean_text(markup_text, steps=["html", "all_whitespace"])  # I think?

citations = get_citations(
    plain_text=plain_text,
    markup_text=markup_text,
    clean_steps=["html", "all_whitespace"]
)
linked_text = annotate_citations(
    plain_text = plain_text,
    source_text = markup_text
    annotations=[[c.span(), "<a>", "</a>"] for c in citations]
)
```

This PR is purely refactoring the ergonomics of the API -- it doesn't change any functionality. (The changes to the tests are just syntactical.) I think it's an improvement, but I recognize that breaking the API is always controversial, so open to discussion. And I think the idea to introduce the `Document` abstraction in the first place is great.

Some other documentation improvement-related PRs to follow soon as well.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -66,7 +66,7 @@ Install eyecite::
 
 Here's a short example of extracting citations and their metadata from text using eyecite's main :code:`get_citations()` function::
 
-    from eyecite import get_citations
+    from eyecite import get_citations, Document
 
     text = """
         Mass. Gen. Laws ch. 1, § 2 (West 1999) (barring ...).
@@ -75,7 +75,8 @@ Here's a short example of extracting citations and their metadata from text usin
         Foo, supra, at 5.
     """
 
-    get_citations(text)
+    document = Document(text)
+    get_citations(document)
 
     # returns:
     [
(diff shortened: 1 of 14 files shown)


flooie:
@mattdahl I need to look at this closer but I think there is some difference between how I would expect a user to generate linked_text 

```
markup_text = "<html goes here...>"
citations = get_citations(
     markup_text=markup_text, clean_steps=["html", "all_whitespace"]
)
document = citations[0].document
linked_text = annotate_citations(
    plain_text=document.plain_text,
    annotations=[(c.span(), "<a>", "</a>") for c in citations],
    source_text=document.markup_text,
    offset_updater=document.plain_to_markup,
)
```

we store a copy of the document so we dont have to generate or regenerate it for making annotations. Although I can imagine annotating looks cleaner here.  but wonder if that makes a difference for you

mattdahl:
Ah I see, I missed that. I think that pattern has two things going on:

1) Creating the document. Currently this happens silently as a byproduct of calling `get_citations()` -- I think it's better for the user to consciously see that happening by instantiating the object himself, so he knows it exists for later.
2) Accessing the document. Currently this is done by grabbing it from `citations[0].document` -- following from (1), this feels a bit hidden to me. The relationship also feels reversed to me -- if anything, a document should have citations, not the other way around? Here, every citation will have a pointer to the *same* document, right?

flooie:
@mattdahl 

I think you are right that it is hidden but I think if I was refactoring I would want this.

```
from eyecite import Document

doc = Document(
    markup_text=opinion_text,
    clean_steps=["html", "all_whitespace"],
)

# or plain text:
# doc = Document(
#     plain_text=opinion_text,
#     clean_steps=["all_whitespace"],
# )

# pull out all the Citation objects
citations = doc.get_citations()

# or perhaps its just doc.citations - because get citations is generated automatically

# build your <a> spans and apply them
markup_annotations = doc.annotate_markup(
    [(c.span(), "<a>", "</a>") for c in citations]
)
```

I could argue that we just pass in your markup or your plaintext with the bool -is_markup =True/false and let it handle the the extra cleanup steps.   But I havent looked at the code in a minute so im not sure if there are any downsides to this pattern.

mattdahl:
Sorry for the slow reply on this. I think I am open to that proposal, but it would effectively switch the entire eyecite API to an OOP paradigm instead of individual function calls. Again, I am open to this (and think it would be an improvement over the current syntax), but are there downsides here we should consider? @mlissner, perhaps you'd like to weigh in?

mlissner:
I don't find the current state of things to be too annoying, but I do like @flooie's simpler version. Is it worth re-doing? Maybe? I would just want to make sure that we document it really well if we're going to do something this drastic. Something to know is that there are forks of our code that re-do eyecite in TypeScript and maybe Rust, I think, and this will surely mess them up, so we better be sure!

mattdahl:
Okay, that's a good point. However, I think *something* must change with the current syntax -- the reason I created this PR was because I was going to update our documentation, but then became confused myself over how things currently worked. (Particularly with the plain_text / markup_text / source_text trifecta.)

I see three main ways to proceed:

1. Do nothing (one extreme).
2. Do some cleanup of the API, but keep the basic functionalist paradigm intact (middle ground; could be this PR or another approach).
3. Objectify everything, a la @flooie's pseudocode (the other extreme).

mlissner:
I'm +1 on a better API and +0 on how it looks. I leave it to you and @flooie to think through. I guess I like his solution pretty well though. :)

PR #307, opened 2026-06-18
feat(tokenizers): relax inter-token whitespace when matching reporter abbreviations

### LINKED ISSUE #305: Relax inter-token whitespace when matching reporter abbreviations
## Summary

eyecite matches reporter abbreviations with **exact** spacing, so the spaced-out print style common in older/official reporters fails to match. `N.Y.S.2d` is in `reporters-db`, but the same citation as printed — `123 N. Y. S. 2d 456` — is not recognized, because the matcher requires the periods and the single space to land exactly where the abbreviation string has them.

Allowing **flexible whitespace between abbreviation tokens** (while keeping periods mandatory) would:

- recover a large class of missed full citations — it is the single biggest cause in the "1b. Missed full citation" bucket of #304;
- make **230** hand-maintained whitespace variations in `reporters-db` redundant and stop the recurring "add a respaced variation" PRs.

> [!NOTE]
> Throughout this issue, gains and costs are reported separately for **whitespace-only** relaxation (the proposal) and **whitespace + periods** relaxation (a more aggressive option that is explicitly out of scope).

This change is mostly localized and feasible: under the **`HyperscanTokenizer`** (what CourtListener — the end consumer — runs in production), the relaxed regex is compiled directly and the fix is essentially the small `_substitute_edition` helper below. The only extra care is for the default `AhocorasickTokenizer`'s literal prefilter, covered in a [subsection](#note-ahocorasicktokenizer-prefilter).

All code links below are pinned to HEAD (`04d82c032ad5fd0f9ab72a61c87110c46ee8f52e`).

## Root cause

Every reporter string is turned into an **exact-match** regex in [`_substitute_edition`](https://github.com/freelawproject/eyecite/blob/04d82c032ad5fd0f9ab72a61c87110c46ee8f52e/eyecite/tokenizers.py#L104-L107):

```python
def _substitute_edition(template, *edition_names):
    """Helper to replace $edition in template with edition_names."""
    edition = "|".join(re.escape(e) for e in edition_names)   # <-- here
    return Template(template).safe_substitute(edition=edition)
```

[`re.escape`](https://github.com/freelawproject/eyecite/blob/04d82c032ad5fd0f9ab72a61c87110c46ee8f52e/eyecite/tokenizers.py#L106) makes:

- periods match a literal period only — `F.R.D.` → `F\.R\.D\.`
- a literal space match **exactly one** space — `App. Div.` → `App\.\ Div\.`

### End-to-end, for `N.Y.S.2d`

1. The edition string is escaped to `N\.Y\.S\.2d`.
2. It is substituted into the default template [`"$volume $reporter,? $page"`](https://github.com/freelawproject/eyecite/blob/04d82c032ad5fd0f9ab72a61c87110c46ee8f52e/eyecite/tokenizers.py#L100), producing (simplified): `(?P<volume>\d+) (?P<reporter>N\.Y\.S\.2d),? (?P<page>\d+)`
3. It is wrapped with non-alphanumeric boundaries by [`nonalphanum_boundaries_re`](https://github.com/freelawproject/eyecite/blob/04d82c032ad5fd0f9ab72a61c87110c46ee8f52e/eyecite/regexes.py#L14-L16).

The result matches `123 N.Y.S.2d 456` but **not** `123 N. Y. S. 2d 456` (the periods are followed by spaces) nor `123 N.Y.S. 2d 456`.

The same path runs for both editions and their variations in [`_add_regexes`](https://github.com/freelawproject/eyecite/blob/04d82c032ad5fd0f9ab72a61c87110c46ee8f52e/eyecite/tokenizers.py#L167-L180).

## Evidence it's a real recall gap

Running the current eyecite on each reporter in isolation, compact vs. spaced:

| Form | Result |
|---|---|
| `260 F.R.D. 38`, `799 N.Y.S.2d 795`, `8 N.Y.3d 204`, `11 Ohio C.C. (n.s.) 553` | ✅ matched |
| `260 F. R. D. 38`, `799 N. Y. S. 2d 795`, `8 N. Y. 3d 204`, `11 Ohio C. C. (N.S.) 553` | ❌ missed |

The reporters are already in `reporters-db`; only the spacing differs.

### Concrete impact on the #304 ground-truth set

These full citations were missed and had to be hand-added (#304 category 1b) for no reason other than inter-token spacing. Each links back to its CourtListener cluster:

| Cluster | Missed (as printed) | Reporter (compact form eyecite already has) |
|---|---|---|
| [103402](https://www.courtlistener.com/opinion/103402/) | `11 Ohio C. C. (N.S.) 553`, `31 Ohio C. C. 270` | `Ohio C.C.` |
| [1722](https://www.courtlistener.com/opinion/1722/) | `260 F. R. D. 38`, `54 F. R. D. 412`, `182 F. R. D. 72` | `F.R.D.` |
| [1722](https://www.courtlistener.com/opinion/1722/) | `799 N. Y. S. 2d 795`, `778 N. Y. S. 2d 147`, `673 N. Y. S. 2d 659`, `665 N. Y. S. 2d 361`, `517 N. Y. S. 2d 764`, `499 N. Y. S. 2d 693` | `N.Y.S.2d` |
| [2621076](https://www.courtlistener.com/opinion/2621076/) | `267 N. Y. S. 2d 564` | `N.Y.S.2d` |

All of these are recovered by **whitespace-only** relaxation. Adding period relaxation recovers **no additional** citations in this set (the remaining 1b misses are spelled-out names like `App. Div.` / `Nebraska` and reporters absent from `reporters-db`, neither of which punctuation relaxation can reach).

## Many spacing/punctuation variations are maintained by hand

`reporters-db` already carries a large number of variations whose only difference from their canonical edition is spacing or punctuation. Counting all of them:

| Variation differs from canonical by… | Count | Made redundant by | Example |
|---|---|---|---|
| whitespace only | **230** | whitespace-only ✅ | `A. 2d` → `A.2d`, `A.D. 3d` → `A.D.3d` |
| periods only | 184 | whitespace + periods | `A2d` → `A.2d`, `AD3d` → `A.D.3d` |
| both | 41 | whitespace + periods | `F 2d` → `F.2d`, `AD 3d` → `A.D.3d` |

A match-time whitespace rule subsumes the **230** whitespace-only entries (in both directions — it doesn't matter whether the variation adds or removes spaces). Whitespace + periods would subsume all 455, but periods are out of scope here (collision risk, below).

These are added by hand, one reporter at a time. Whitespace-variant PRs do happen — e.g. freelawproject/reporters-db#185 (`App.D.C.`→`App. D.C.`, `Cl.Ct.`→`Cl. Ct.`, `Mo.App.`→`Mo. App.`), #149 (`N. C.`→`N.C.`), #141 (`Abb.N.Cas.`→`Abb. N. Cas.`), #132 (`F.App'x`→`F. App'x`) — but in fairness the recent variation PRs lean toward *period* fixes (e.g. #202 `AD3d`→`A.D.3d`, #201 `NY2d`→`N.Y.2d`, #205 `F3d`→`F.3d`), which this proposal deliberately does not tackle. So the primary justification here is the recall gain in #304, with the 230 redundant whitespace variations a secondary cleanup.

## Proposed change

Relax **whitespace only** (keep periods mandatory) at the single chokepoint in [`_substitute_edition`](https://github.com/freelawproject/eyecite/blob/04d82c032ad5fd0f9ab72a61c87110c46ee8f52e/eyecite/tokenizers.py#L104-L107). After escaping, make whitespace flexible:

```python
def _relax_ws(escaped: str) -> str:
    # "N.Y.S.2d" should also match "N. Y. S. 2d";
    # "App. Div." should also match "App.Div." / "App.  Div."
    escaped = escaped.replace(r"\.", r"\.\s*")        # period -> period + optional ws
    escaped = re.sub(r"(?:\\ )+", r"\\s+", escaped)   # literal space(s) -> \s+
    return escaped

def _substitute_edition(template, *edition_names):
    edition = "|".join(_relax_ws(re.escape(e)) for e in edition_names)
    return Template(template).safe_substitute(edition=edition)
```

Notes:

- `\.\s*` (period + optional whitespace) covers `N. Y. S. 2d` and `N.Y.S. 2d`; `\s+` for existing spaces keeps multi-word reporters intact (`Ohio␣C.C.`).
- Consider `[ \t]` or an explicit `[\s ]` class rather than `\s` to avoid matching across newlines / to include non-breaking spaces — pick what matches eyecite's text-cleaning assumptions.
- Deliberately **whitespace only**. Making periods optional too (`AD3d` → `A.D.3d`) is more powerful but considerably riskier — see collisions.

### `HyperscanTokenizer` — production path, works directly

CourtListener runs the [`HyperscanTokenizer`](https://github.com/freelawproject/eyecite/blob/04d82c032ad5fd0f9ab72a61c87110c46ee8f52e/eyecite/tokenizers.py#L472-L494), which compiles the actual extractor regexes and scans in a single pass. It has no literal-string prefilter, so it picks up the relaxed pattern for free — no extra work beyond the `_substitute_edition` change. (One thing to verify: hyperscan accepts the added `\s*` / `\s+`, which it should as standard syntax.)

The plain [`Tokenizer`](https://github.com/freelawproject/eyecite/blob/04d82c032ad5fd0f9ab72a61c87110c46ee8f52e/eyecite/tokenizers.py#L325) has no prefilter either and also works directly.

So for the production target this is **mostly a localized, low-risk change**: the helper above plus tests.

<a name="note-ahocorasicktokenizer-prefilter"></a>
### Note: `AhocorasickTokenizer` prefilter

The default tokenizer is [`AhocorasickTokenizer`](https://github.com/freelawproject/eyecite/blob/04d82c032ad5fd0f9ab72a61c87110c46ee8f52e/eyecite/tokenizers.py#L611), which uses a literal-string prefilter and needs a small adjustment (not relevant to CourtListener, but worth getting right for library users). Two coupled spots:

1. [`_add_regex`](https://github.com/freelawproject/eyecite/blob/04d82c032ad5fd0f9ab72a61c87110c46ee8f52e/eyecite/tokenizers.py#L145-L148) registers prefilter strings only when `re.escape(reporters[0]) in regex`. After relaxation the literal `N\.Y\.S\.2d` is no longer a substring, so this becomes `False` and the `strings` set is left empty.
2. [`AhocorasickTokenizer`](https://github.com/freelawproject/eyecite/blob/04d82c032ad5fd0f9ab72a61c87110c46ee8f52e/eyecite/tokenizers.py#L412-L445) runs extractors with no `strings` on *every* input, so empty `strings` would push reporter extractors into the always-run set and erode the optimization. (And a compact prefilter string wouldn't fire on `N. Y. S. 2d` anyway, since Aho-Corasick matches literal substrings.)

Simplest fix that preserves the optimization: build the automaton from the compact strings but run it against a **whitespace-stripped copy of the text used only for extractor selection**, while still matching the real relaxed regex against the original text for spans/offsets. This keeps selection a superset of real matches. Alternatively, scope relaxation to the single-letter+period reporters (where the 304 misses cluster) to bound the blast radius.

## Collision analysis

Methodology: normalize every edition **and** variation string across all 1,236 reporters in `reporters-db`, then count **distinct reporter pairs** that become indistinguishable under each relaxation (a pair already ambiguous today doesn't count as new).

| Relaxation | New ambiguous reporter pairs |
|---|---|
| Whitespace only (periods kept) | **+5** |
| Whitespace **+** periods | +30 |

The 5 new whitespace-only collisions, all between rare/historical reporters:

```
Am. Neg. Ca.        <> Am. Neg. Cases      # same reporter (harmless)
Dem. Sur.           <> Dem. Surr.          # same reporter (harmless)
City Hall Rec. (NY) <> N. Y. City H. Rec.  # same NY reporter (harmless)
Pa. C.              <> Pa. Commw.          # both PA intermediate; worth a look
Dev. & Bat.         <> N.C.                # worth review (arises from a variation)
```

Four of five are effectively the same reporter.

Making **periods** optional too is far worse — **+30** new ambiguous pairs vs. baseline (25 of them introduced specifically by dropping periods, on top of the 5 whitespace ones). Unlike the whitespace collisions, many of these merge *genuinely distinct* reporters, which is why this proposal stops at whitespace. Examples by type (key → reporters it would conflate):

**Neutral-citation court codes collapsing into state reporters** (the most concerning class — periods are the only thing distinguishing them):

```
mt  -> {MT (Montana neutral),       Mont.}
la  -> {LA (Louisiana neutral),     La.}
nd  -> {ND (North Dakota neutral),  N.D.}
nm  -> {NM (New Mexico neutral),    N.M.}
sd  -> {SD (South Dakota neutral),  S.D.}
pa  -> {PA (Pennsylvania neutral),  Pa.}
vt  -> {VT (Vermont neutral),       Vt.}
wy  -> {WY (Wyoming neutral),       Wyo.,  Va. Ch. Dec.}
```

**Distinct reporters merged, including three-way collisions:**

```
pac -> {P. (Pacific),  Pa. C.,  Pa. Commw.}
sc  -> {S. Ct. (Supreme Court Reporter),  S.C. (South Carolina),  Scam.}
cit -> {CIT,  Ct. Int'l Trade}
clr -> {Cl. Ch.,  Conn. L. Rptr.}
vr  -> {Vroom,  Vt.}
mon -> {B. Mon.,  Monag.,  Mont.,  T.B. Mon.}
w   -> {Wash.,  Wend.,  Wis.,  Wyo.}
```

**Reporter vs. sub-series merged:**

```
calappsupp   -> {Cal. App.,      Cal. App. Supp.}
calapp2dsupp -> {Cal. App. 2d,   Cal. App. Supp. 2d}
black        -> {Black,          Blackf.}
mass(cush)   -> {Cush.,          Mass.}
```

The neutral-vs-regional class is the strongest argument against period relaxation: collapsing e.g. `MT` into `Mont.` would silently mis-resolve modern Montana neutral citations to the Montana Reports.

<details>
<summary>Reproduce the collision counts</summary>

```python
import json, re
from itertools import combinations
# reporters_db/data/reporters.json
d = json.load(open("reporters.json"))
strings = {}
for _, lst in d.items():
    for rep in lst:
        for ed in rep.get("editions", {}):
            strings.setdefault(ed, set()).add(ed)
        for var, canon in rep.get("variations", {}).items():
            strings.setdefault(var, set()).add(canon)

def pairs(normfn):
    m = {}
    for s, canons in strings.items():
        m.setdefault(normfn(s), set()).update(canons)
    out = set()
    for v in m.values():
        if len(v) > 1:
            out |= set(combinations(sorted(v), 2))
    return out

base = pairs(lambda s: s.lower())
ws   = pairs(lambda s: re.sub(r"\s+", "", s).lower())
both = pairs(lambda s: re.sub(r"[\s.]", "", s).lower())
print("new ws-only pairs:", len(ws - base))
print("new ws+period pairs:", len(both - base))
print(sorted(ws - base))
```
</details>

## Scope / non-goals

This fixes spaced **abbreviations** only. It does **not** address the following, which remain separate `reporters-db` data additions:

- spelled-out reporter names that aren't registered variants — `App. Div.` (≠ `A.D.`), `Nebraska` (≠ `Neb.`), `California` (≠ `Cal.`)
- reporters absent from `reporters-db` entirely — e.g. `Clark & Fin.`, `Eng. Rep.`
- making **periods** optional (`NY2d`, `AD3d`) — higher collision risk; out of scope

## Suggested validation

- Unit tests for spaced forms: `123 N. Y. S. 2d 456`, `260 F. R. D. 38`, `11 Ohio C. C. (N. S.) 553`, `8 N. Y. 3d 204`, and that compact forms still match.
- A short-cite case (e.g. `N. Y. S. 2d, at 797`) to confirm [`short_cite_re`](https://github.com/freelawproject/eyecite/blob/04d82c032ad5fd0f9ab72a61c87110c46ee8f52e/eyecite/regexes.py#L19-L23) still composes with the relaxed pattern.
- Re-run against the #304 ground-truth set to measure recall gain and confirm no precision regression.
- Benchmark the `HyperscanTokenizer` (CourtListener's path) for throughput; spot check the `AhocorasickTokenizer` once its prefilter adjustment is in place.
- Confirm the 5 collision pairs behave acceptably (most are the same reporter).

grossir:
Fixes #305

- Add _relax_ws helper: after re.escape, periods stay mandatory but inter-token whitespace becomes optional (N.Y.S.2d also matches N. Y. S. 2d).
- Apply at the single chokepoint _substitute_edition, so all reporter regexes (plain Tokenizer + production HyperscanTokenizer) pick it up for free.
- Fix AhocorasickTokenizer prefilter: match whitespace-stripped reporter strings against a whitespace-stripped copy of the text (selection only; real spans still come from the regex).
- Update _add_regex have_strings check to look for the relaxed form so prefilter strings still register.
- Scope: whitespace only — periods stay mandatory (avoids the +30 collision risk of dropping periods, e.g. MT→Mont.); +5 whitespace collisions, all rare/duplicate reporters.
- Tests: new test_relaxed_reporter_whitespace (single-letter, multi-word, partial, compact forms across all 3 tokenizers); updated test_reporter_tokenizer (U. S. now also an exact match) and test_extractor_filter.

### DIFF
--- a/tests/test_FindTest.py
+++ b/tests/test_FindTest.py
@@ -878,6 +878,18 @@ class FindTest(TestCase):
             ('799 N.Y.S.2d 795',
              [case_citation(volume='799', page='795',
                             reporter='N.Y.S.2d')]),
+            # Short cite with a spaced reporter, which goes through the
+            # separate short_cite_re path. "N. Y. S. 2d" is not a registered
+            # variation, so it matches only via the whitespace relaxation --
+            # unlike the "U. S." short cites elsewhere in this file, which
+            # still pass through their reporters-db variation. Once those
+            # whitespace-only variations are removed (reporters-db #261),
+            # this relaxation becomes the only path for spaced short cites,
+            # so guard that short_cite_re composes with it.
+            ('See Adarand, 799 N. Y. S. 2d, at 797',
+             [case_citation(volume='799', page='797', reporter='N.Y.S.2d',
+                            reporter_found='N. Y. S. 2d', short=True,
+                            metadata={'antecedent_guess': 'Adarand'})]),
         )
         # fmt: on
         self.run_test_pairs(test_pairs, "Relaxed reporter whitespace")
--- a/tests/test_FindTest.py
+++ b/tests/test_FindTest.py
@@ -890,6 +890,17 @@ class FindTest(TestCase):
              [case_citation(volume='799', page='797', reporter='N.Y.S.2d',
                             reporter_found='N. Y. S. 2d', short=True,
                             metadata={'antecedent_guess': 'Adarand'})]),
+            # Law citations (laws.json) and journal citations (journals.json)
+            # build their regexes through the same chokepoint, so the
+            # relaxation applies to them too. Spaced "U. S. C." / "F. B. I. S."
+            # are not registered variations, so these only match via it.
+            ('18 U. S. C. § 4241',
+             [law_citation('18 U. S. C. § 4241', reporter='U.S.C.',
+                           reporter_found='U. S. C.',
+                           groups={'title': '18', 'section': '4241'})]),
+            ('1 F. B. I. S. 1',
+             [journal_citation('1 F. B. I. S. 1', reporter='F.B.I.S.',
+                               reporter_found='F. B. I. S.')]),
         )
         # fmt: on
         self.run_test_pairs(test_pairs, "Relaxed reporter whitespace")
--- a/tests/test_FindTest.py
+++ b/tests/test_FindTest.py
@@ -841,6 +841,47 @@ class FindTest(TestCase):
         # fmt: on
         self.run_test_pairs(test_pairs, "Citation extraction")
 
+    def test_relaxed_reporter_whitespace(self):
+        """Do we match reporters printed with extra inter-token whitespace?
+
+        Reporter abbreviations are stored compactly in reporters-db (e.g.
+        "N.Y.S.2d"), but the older/official print style spaces the tokens out
+        ("N. Y. S. 2d"). Whitespace between tokens is relaxed at match time
+        while periods stay mandatory. See #305.
+        """
+        # fmt: off
+        test_pairs = (
+            # Spaced single-letter reporters
+            ('123 N. Y. S. 2d 456',
+             [case_citation(volume='123', page='456',
+                            reporter='N.Y.S.2d',
+                            reporter_found='N. Y. S. 2d')]),
+            ('260 F. R. D. 38',
+             [case_citation(volume='260', page='38',
+                            reporter='F.R.D.',
+                            reporter_found='F. R. D.')]),
+            ('8 N. Y. 3d 204',
+             [case_citation(volume='8', page='204',
+                            reporter='N.Y.3d',
+                            reporter_found='N. Y. 3d')]),
+            # Partially spaced (space only after the final period)
+            ('799 N.Y.S. 2d 795',
+             [case_citation(volume='799', page='795',
+                            reporter='N.Y.S.2d',
+                            reporter_found='N.Y.S. 2d')]),
+            # Multi-word reporter with spaced tokens
+            ('31 Ohio C. C. 270',
+             [case_citation(volume='31', page='270',
+                            reporter='Ohio C.C.',
+                            reporter_found='Ohio C. C.')]),
+            # Compact forms still match unchanged
+            ('799 N.Y.S.2d 795',
+             [case_citation(volume='799', page='795',
+                            reporter='N.Y.S.2d')]),
+        )
+        # fmt: on
+        self.run_test_pairs(test_pairs, "Relaxed reporter whitespace")
+
     def test_find_law_citations(self):
         """Can we find citations from laws.json?"""
         # fmt: off
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 The following changes are not yet released, but are code complete:
 
 Features:
--
+- Relax inter-token whitespace when matching reporter abbreviations, so spaced-out print forms like `123 N. Y. S. 2d 456` match the compact `N.Y.S.2d` in reporters-db. Periods stay mandatory. #305
 
 Changes:
 -
--- a/eyecite/tokenizers.py
+++ b/eyecite/tokenizers.py
@@ -92,6 +92,44 @@ def token_is_from_nominative_reporter(token: Token) -> bool:
     return name in NOMINATIVE_REPORTER_NAMES
 
 
+_WHITESPACE_RE = re.compile(r"\s+")
+
+
+def _relax_ws(escaped: str) -> str:
+    """Relax inter-token whitespace in an escaped reporter abbreviation.
+
+    ``re.escape`` turns a reporter string into an exact-match pattern in which
+    each period matches a literal period and each space matches exactly one
+    space. That makes the spaced-out print style common in older/official
+    reporters fail to match: ``N.Y.S.2d`` is in reporters-db, but the same
+    citation as printed -- ``N. Y. S. 2d`` -- is not recognized.
+
+    This makes the whitespace *between* abbreviation tokens flexible while
+    keeping periods mandatory, so a single canonical reporter string matches
+    all of its respaced forms (e.g. ``N.Y.S.2d`` also matches ``N. Y. S. 2d``
+    and ``N.Y.S. 2d``; ``Ohio C.C.`` also matches ``Ohio  C. C.``).
+
+    See https://github.com/freelawproject/eyecite/issues/305.
+    """
+    # period -> period followed by optional whitespace
+    escaped = escaped.replace(r"\.", r"\.\s*")
+    # one or more literal spaces -> optional whitespace.
+    # (`re.escape` escapes a space as "\ "; match it with or without the
+    # backslash so this is robust across Python versions.)
+    escaped = re.sub(r"(?:\\?\ )+", r"\\s*", escaped)
+    return escaped
+
+
+def _strip_ws(text: str) -> str:
+    """Remove all whitespace from ``text``.
+
+    Used by :class:`AhocorasickTokenizer` to match its whitespace-insensitive
+    reporter prefilter strings against text that may use a spaced-out print
+    style. See :func:`_relax_ws`.
+    """
+    return _WHITESPACE_RE.sub("", text)
+
+
 def _populate_reporter_extractors():
     """Populate EXTRACTORS and EDITIONS_LOOKUP."""
 
@@ -103,7 +141,7 @@ def _populate_reporter_extractors():
 
     def _substitute_edition(template, *edition_names):
         """Helper to replace $edition in template with edition_names."""
-        edition = "|".join(re.escape(e) for e in edition_names)
+        edition = "|".join(_relax_ws(re.escape(e)) for e in edition_names)
         return Template(template).safe_substitute(edition=edition)
 
     # Extractors step one: add an extractor for each reporter string
@@ -143,7 +181,9 @@ def _populate_reporter_extractors():
         editions_by_regex[regex][kind].append(edition)
 
         # add strings
-        have_strings = re.escape(reporters[0]) in regex
+        # The reporter is embedded in `regex` with relaxed inter-token
+        # whitespace (see _relax_ws), so check for the relaxed form.
+        have_strings = _relax_ws(re.escape(reporters[0])) in regex
         if have_strings:
             editions_by_regex[regex]["strings"].update(reporters)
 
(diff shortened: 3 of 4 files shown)


grossir:
About the benchmark, remember Gains are actually losses

About the losses: All 6 are a bare fragment in the base being absorbed into a now-complete citation in the PR:

```
  ┌─────────┬───────────────┬────────────────────────────────────────────┐
  │   id    │    "Lost"     │          Actually gained instead           │
  ├─────────┼───────────────┼────────────────────────────────────────────┤
  │ 5160500 │ §             │ Kan.Stat. Ann. § 59-2103 (+2 journals)     │
  ├─────────┼───────────────┼────────────────────────────────────────────┤
  │ 2357843 │ §, §§         │ 7× full D.C.Code § cites                   │
  ├─────────┼───────────────┼────────────────────────────────────────────┤
  │ 2647974 │ §             │ Idaho\nCode § 18-1506                      │
  ├─────────┼───────────────┼────────────────────────────────────────────┤
  │ 603885  │ §             │ Ark.Code Ann. § 4-9-106                    │
  ├─────────┼───────────────┼────────────────────────────────────────────┤
  │ 1308185 │ Thompson, 452 │ 452 FSupp. 165 (reference→full-cite shift) │
  └─────────┴───────────────┴────────────────────────────────────────────┘

```

tests/test_FindTest.py
```diff
@@ -841,6 +841,47 @@ def test_find_citations(self):
...
+            # Multi-word reporter with spaced tokens
+            ('31 Ohio C. C. 270',
+             [case_citation(volume='31', page='270',
+                            reporter='Ohio C.C.',
+                            reporter_found='Ohio C. C.')]),
+            # Compact forms still match unchanged
+            ('799 N.Y.S.2d 795',
+             [case_citation(volume='799', page='795',
+                            reporter='N.Y.S.2d')]),
+        )
```
quevon24, line 904: Consider adding a spaced short cite here. The full-cite forms are well covered, but short cites go through short_cite_re (a separate path) and nothing currently, guards that the relaxation composes with it.

The U. S. short cites elsewhere in this file pass only because U. S. is a registered variation in reporters-db, so they wouldn't catch a regression here. A reporter whose spaced form is not a variation (e.g. N.Y.S.2d) does:

```suggestion
   # Short cite with a spaced reporter. Unlike "U. S.", the spaced
   # "N. Y. S. 2d" is not a registered variation, so this only matches
   # via the relaxation. Guards that short_cite_re composes with it.
   ('See Adarand, 799 N. Y. S. 2d, at 797',
    [case_citation(volume='799', page='797', reporter='N.Y.S.2d',
                   reporter_found='N. Y. S. 2d', short=True,
                   metadata={'antecedent_guess': 'Adarand'})]),
)
```

quevon24 (approved):
Looks good, and the suite passes locally across all three tokenizers (plain, Aho-Corasick, Hyperscan). I only have one non-blocking suggestion

tests/test_FindTest.py
grossir: Thanks, I added tests for this and for other citation types; FullLawCitation,FullJournalCitation

Issue #212, opened 2025-02-14
Improve benchmark


grossir:
First, the readability can be improved. Branches are named "branch1", "branch2"; but which is "main" and which is the PR?

Second, we could return some statistics already implied in the processing
In this recent [PR](https://github.com/freelawproject/eyecite/pull/203), there was a noticeable processing time increment, which is an important metric for the changes

Also, with the introduction of ReferenceCitations we get the possibility of overlapping citations; it would be interesting to return the citation type, besides the citation type, in the JSON itself

About processing time calculation, from the results JSON themselves 
```python
url1 = "https://raw.githubusercontent.com/freelawproject/eyecite/artifacts/203/results/8981703e7cc27067adcb39f66346dc62248974cf.json"
url2 = "https://raw.githubusercontent.com/freelawproject/eyecite/artifacts/203/results/bb9ca00f64c5aa47d0eb85a16e38bc03a6bf0b61.json"

def get_time_stats(url):
    import requests
    import statistics
    
    jason = requests.get(url).json()
    prev_start = jason[0]['time']
    actual_times = [prev_start]
    for item in jason[1:]:
        actual_times.append(item['time'] - prev_start)
    
    print("Mean: ", sum(actual_times)/len(actual_times))
    print("Median: ", statistics.median(actual_times))
    print("Sample size: ", len(actual_times))

get_time_stats(url1)
get_time_stats(url2)
```

Yields
```
Mean:  35.24138639922976
Median:  37.765769999999996
Sample size:  779

Mean:  20.421772406931975
Median:  21.761733
Sample size:  779
```

grossir:
Another improvement: benchmark annotation performance

On a couple of recent PRs, the peformance footprint of changes (new cleaning step, new span updater) was invisible when they were on the annotation [step](https://github.com/freelawproject/eyecite/pull/269); but became evident when moved to the find [step](https://github.com/freelawproject/eyecite/pull/270)

grossir:
## Bug: the report's Gain/Loss columns are inverted

Related to point #1 above ("which branch is `main` and which is the PR?") — this ambiguity currently produces an actual **correctness bug**: the report swaps Gains and Losses on every PR.

Concrete example, PR #307 (relaxes reporter whitespace, so it should *gain* spaced forms like `111 N. Y. S. 2d 500`). The bot reported:

> There were 6 gains and 238 losses.

[...]

PR #309, opened 2026-06-18
fix(benchmark): correct inverted gains/losses and broken report table

grossir:
Fixes the benchmark report bugs surfaced while reviewing #307. Refs #212.

## Problems

1. **Gains and Losses were inverted on every eyecite PR.** The diff logic assumes `--branches[0]` is the baseline, but the eyecite workflow passes `--branches <PR> <main>` (PR first), while the reporters-db workflow passes `--branches original update` (base first). So one path was always backwards. On #307 the report read "6 gains and 238 losses" when it was actually **238 gains / 6 losses**.
2. **The Markdown table broke** whenever a citation token contained an embedded newline (e.g. `163 Ohio\nSt.3d 242`) — a blank line inside a table ends it on GitHub, so rows silently disappeared.
3. **Documents were compared by `zip` position**, so a single skipped/reordered document would misalign every later comparison.
4. The "only displaying the first 50" notice keyed off **column widths**, not the change count, and was off-by-one.

## Changes

- Orient gains/losses with an explicit `--base` flag. It defaults to the first branch, so the reporters-db ordering keeps working; the eyecite workflow now passes `--base <main>`.
- Sanitize Markdown table cells: collapse whitespace runs (embedded newlines included) and escape pipes. The CSV artifact keeps the raw token text.
- Compare documents by `id` (union of both runs) instead of by position.
- Truncation notice uses the real change count and emits exactly `MAX_ROWS_IN_MD` rows.
- Add a net citation-count summary (base vs PR, with delta).
- Clearer base/PR labels; migrate deprecated `::set-output` → `$GITHUB_OUTPUT`.

## Verification

Exercised `generate_time_chart` + `write_report` with synthetic base/PR JSONs:
- correct gain/loss direction;
- ids present on only one side handled (which `zip` would mispair);
- embedded-newline + pipe cells render cleanly while the CSV stays raw;
- `--base` orientation correct for both workflow argument orders + the no-`--base` default;
- truncation message shows the true count with exactly 50 rows.

YAML parses; `ruff check`/`format` clean; existing test suite passes.

## Deferred (follow-ups, noted on #212)

Per-doc timing statistics, citation type/overlap in the JSON, Hyperscan-path benchmarking, and multiset counting.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -46,12 +46,15 @@ jobs:
           echo "$VIRTUAL_ENV/bin" >> $GITHUB_PATH
           echo "VIRTUAL_ENV=$VIRTUAL_ENV" >> $GITHUB_ENV
 
+      # branch1 = the PR (the repo is checked out at the PR ref here, so
+      # `env.main` resolves to the PR head sha). The baseline (main) is
+      # captured later as branch2, after we check main out.
       - name: Setup variables I
         id: branch1
         run: |
           echo ${{ github.event.issue.pull_request }}
-          echo "::set-output name=filepath::benchmark/${{ env.main }}.json"
-          echo "::set-output name=hash::${{ env.main }}"
+          echo "filepath=benchmark/${{ env.main }}.json" >> "$GITHUB_OUTPUT"
+          echo "hash=${{ env.main }}" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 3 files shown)


quevon24 (approved):
LGTM. Clean, well-targeted fixes. Approving.

PR #319, opened 2026-07-01
fix: de-duplicate identical editions in all_editions (#317)

### LINKED ISSUE #317: all_editions contains duplicate (identical) editions for whitespace-spaced reporters since 2.7.7
## Summary

Since v2.7.7, `get_citations()` returns `FullCaseCitation` objects whose `all_editions` list contains the **same `Edition` object twice** for any reporter that has a whitespace-only variation in reporters-db (e.g. `F.3d`, `F.2d`, `U.S.`, `Cal. 4th`). The duplicate is the identical object — not two distinct editions — so it is a spurious duplicate rather than genuine ambiguity.

This was introduced by the whitespace-relaxation change in #305 / #307.

## Reproduction

```python

c = get_citations("238 F.3d 273")[0]
print(len(c.all_editions))            # 2  (was 1 in 2.7.6)
print(len(set(c.all_editions)))       # 1
print(c.all_editions[0] is c.all_editions[1])  # True
```

```
Edition(reporter=Reporter(short_name='F.', name='Federal Reporter', cite_type='federal', ...),
        short_name='F.3d', start=datetime.datetime(1993, 1, 1, 0, 0), end=None)
```

Affected across most multi-series case reporters:

| citation | `len(all_editions)` 2.7.6 | 2.7.7 | distinct |
|---|---|---|---|
| `238 F.3d 273` | 1 | **2** | 1 |
| `410 U.S. 113` | 1 | **2** | 1 |
| `123 F.2d 456` | 1 | **2** | 1 |
| `5 Cal. 4th 1` | 1 | **2** | 1 |
| `72 Soc.Sec.Rep.Serv. 318` (no ws-only variation) | 1 | 1 | 1 |

Confirmed it's eyecite, not reporters-db: `eyecite==2.7.7` + `reporters-db==3.2.65` reproduces; `eyecite==2.7.6` + `reporters-db==3.2.66` does not.

## Root cause

In `_populate_reporter_extractors`, `_add_regexes` builds one extractor regex for the canonical edition name and a second one for the edition's variations:

```python
regex = _substitute_edition(regex_template, edition_name)      # "F.3d"  -> F\.\s*3d
_add_regex("editions", [edition_name], edition, regex)
if variations:
    regex = _substitute_edition(regex_template, *variations)   # "F. 3d" -> F\.\s*\s*3d
    _add_regex("variations", variations, edition, regex)
```

After `_relax_ws` makes periods absorb adjacent whitespace, the canonical regex (`F\.\s*3d`) and the variation regex derived from the whitespace-only variation `F. 3d` (`F\.\s*\s*3d`) **both match the plain text `F.3d`**. They're keyed separately in `editions_by_regex`, so two extractors fire on the same span, and each appends the same `Edition` object — yielding the duplicate in `all_editions`.

The whitespace-only variations (`F. 3d`, `U. S.`, etc.) are now redundant precisely because of the relaxation feature; removing them from reporters-db collapses `all_editions` back to 1 (tracked in freelawproject/reporters-db#261). But eyecite arguably shouldn't surface the same edition twice regardless of the source data.

## Impact

Downstream code that uses `len(citation.all_editions)` to detect genuine reporter ambiguity now gets false positives. In CourtListener this dropped valid federal citations in `parse_citations` and broke CI on the routine FLP dependency bump (freelawproject/courtlistener#7500).

## Suggested fix

De-duplicate identical editions when assembling a citation's `all_editions` (or when merging overlapping extractor matches over the same span), so a single canonical citation never lists the same `Edition` more than once. This is robust even where reporters-db retains overlapping variations.

## Related

- Introduced by #305 / #307 (inter-token whitespace relaxation)
- reporters-db cleanup of redundant whitespace-only variations: freelawproject/reporters-db#261
- Surfaced in CourtListener CI: freelawproject/courtlistener#7500

grossir:
## What

Fixes #317. `get_citations()` was returning the same `Edition` object twice in `all_editions`, e.g.:

```python
c = get_citations("238 F.3d 273")[0]
len(c.all_editions)                     # 2 (was 1 pre-#305)
len(set(c.all_editions))                # 1  -> identical objects
c.all_editions[0] is c.all_editions[1]  # True
```

## Root cause

The whitespace relaxation from #305 turns each period/space in a reporter regex into `\s*`. For a reporter with a whitespace-only variation like `F. 3d`, the variation regex becomes `F\.\s*\s*3d`, which also matches the zero-space text `F.3d`. So two extractors fire on the same span — the *exact* one and the *variation* one — and both register the **same** `Edition` object (one into `exact_editions`, one into `variation_editions`). `CitationToken.merge` only deduped *within* each list, never across them, so `all_editions = exact + variation` carried the duplicate. The `S.W.2d` custom regex hit the same bug independently (its regex omits `$edition`), predating #305.

## Fix

- New `_dedupe_editions()` helper: order-preservingly dedupes each list and drops any variation edition already present as an exact match.
- Routed `ResourceCitation.__post_init__`, `CitationToken.__post_init__`, and `CitationToken.merge` through it.
- Genuinely distinct editions (real ambiguity) are preserved.

## Impact

Fixes downstream false positives in callers that use `len(citation.all_editions)` to detect reporter ambiguity (e.g. CourtListener's `update_casenames_wl_dataset`, which was skipping every whitespace-variation reporter as "unable to disambiguate").

Hash note: this changes the hash of *law and journal* citations that previously carried a duplicate edition. Case citations are unaffected — `CaseCitation.__hash__` never depended on `all_editions`. Resolution grouping within a run is unchanged; only absolute hash values (never persisted) differ across versions.

## Tests

- New `test_no_duplicate_editions` — covers `F.3d`, `U.S.`, the `S.W.2d` collapse, and `B.R.` (genuine ambiguity must survive dedup).
- Updated `test_reporter_tokenizer` assertion.
- Full suite: 54 tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,12 @@ Changes:
 - CI: bump GitHub Actions to Node 24 runtimes, clearing the directly-fixable Node 20 deprecation warnings
 
 Fixes:
--
+- De-duplicate identical editions in `all_editions` (and across
+  `exact_editions`/`variation_editions`) so a single canonical citation never
+  lists the same `Edition` twice. The whitespace relaxation in #305 made every
+  reporter with a whitespace-only variation surface a spurious duplicate, and
+  the `S.W.2d` custom regex did so even before #305; distinct editions
+  (genuine ambiguity) are preserved. #317
 
 ## Current
 
--- a/eyecite/models.py
+++ b/eyecite/models.py
@@ -59,6 +59,28 @@ class Edition:
         )
 
 
+def _dedupe_editions(
+    exact: Sequence["Edition"], variation: Sequence["Edition"]
+) -> tuple[tuple["Edition", ...], tuple["Edition", ...]]:
+    """De-duplicate editions while preserving order and genuine ambiguity.
+
+    A single ``Edition`` can be reached both as an exact match and as a
+    variation -- e.g. for reporters with whitespace-only variations that the
+    relaxation in #305 makes redundant, or for the ``S.W.2d`` custom regex that
+    omits ``$edition`` so its canonical and variation forms are identical. That
+    surfaces the same object twice in ``all_editions`` (a spurious duplicate,
+    not genuine ambiguity). Dedupe each list, and drop variation editions that
+    are already exact matches, so distinct editions are preserved but a single
+    edition is never listed more than once.
+    """
+    exact = tuple(dict.fromkeys(exact))
+    seen = set(exact)
+    variation = tuple(
+        e for e in dict.fromkeys(variation) if e not in seen
+    )
+    return exact, variation
+
+
 @dataclass(eq=False, unsafe_hash=False)
 class CitationBase:
     """Base class for objects returned by `eyecite.find.get_citations`. We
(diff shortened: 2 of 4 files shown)


quevon24 (approved):
LGTM, tests pass

PR #328, opened 2026-07-31
Fix get_citations(markup_text=...) crashing without an "html" clean step

renatodvc:
Calling the documented markup path without `clean_steps` raises `TypeError: argument of type 'NoneType' is not iterable` (`clean_steps` defaults to `None`), **and** passing a list without `"html"` raises `TypeError: 'str' object cannot be interpreted as an integer`, because `Document.__post_init__` calls `list.insert()` with its arguments reversed:

```
get_citations(markup_text="<p><i>1 U.S. 1</i></p>")   # -> TypeError
get_citations(markup_text="<p><i>1 U.S. 1</i></p>", clean_steps=["all_whitespace"])  #  -> TypeError
```

The intended fallback (prepend "html" and log a warning) was unreachable. This normalizes `clean_steps` to a list and prepends with the correct argument order, so the fallback now works as designed. No behavior change for callers that already pass "html".

Adds a regression test covering all three cases (omitted, missing "html", already present) and a `CHANGES.md` entry.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,7 @@ Changes:
 -
 
 Fixes:
-- Fix 2 `TypeError`s raised in `get_citations(markup_text=...)` when 
+- Fix 2 `TypeError`s raised in `get_citations(markup_text=...)` when
   `clean_steps` was omitted or lacked `"html"`. The documented fallback that
   prepends the `html` step was both unreachable and broken.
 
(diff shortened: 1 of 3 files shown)


mlissner:
@claude review

mlissner:
To you, @flooie, for triage.

grossir (approved):
@renatodvc thanks for the contribution. This looks good to merge but it's failing the Lint workflow and we need you to sign the CLA

Issue #326, opened 2026-07-16
Bare date parenthetical (e.g. "(May 2, 1999)") mis-captures the month as publisher


grossir:
Parent: freelawproject/eyecite#322

### Summary
When a law citation's parenthetical is a **bare date with no publisher** — e.g. `(May 2, 1999)` — eyecite captures the **month name into the `publisher` field** and leaves `month` empty. The metadata is silently wrong: `publisher` holds `"May"`, `month` is `None`.

This is a regression against documented behavior: `(May 2, 1999)` is one of the forms the code comment above the regex explicitly lists as supported.

```python
# regexes.py, above POST_LAW_CITATION_REGEX:
# statutory and regulatory cites may have publishers and dates after them, like
#  (West), (West 1999), (Lexis Jun. 2018), (1999), or (May 2, 1999),
```

### Minimal reproduction (eyecite 2.7.8)
```python
from eyecite import get_citations

for t in [
    "18 U.S.C. § 1 (May 2, 1999)",        # bare date -> month mis-captured as publisher
    "18 U.S.C. § 1 (Jan. 1, 2012)",       # same
    "18 U.S.C. § 1 (West Jan. 1, 2012)",  # control: publisher present -> month is correct
]:
    for c in get_citations(t):
        m = c.metadata
        print(f"{t!r:36} -> publisher={getattr(m,'publisher',None)!r} "
              f"month={getattr(m,'month',None)!r} day={getattr(m,'day',None)!r} "
              f"year={getattr(m,'year',None)!r}")
```
Output:
```
'18 U.S.C. § 1 (May 2, 1999)'       -> publisher='May'  month=None   day='2' year='1999'
'18 U.S.C. § 1 (Jan. 1, 2012)'      -> publisher='Jan.' month=None   day='1' year='2012'
'18 U.S.C. § 1 (West Jan. 1, 2012)' -> publisher='West' month='Jan.' day='1' year='2012'
```
Expected for `(May 2, 1999)`: `publisher=None, month='May', day='2', year='1999'`.

### Root cause
In `POST_LAW_CITATION_REGEX` the optional publisher group precedes the optional month:
```
(?P<publisher>[A-Z][a-z]+\.?(?:\ Supp\.)?)?   # <- greedily matches a leading "May"/"Jan."
\ ?
(?:{MONTH_REGEX}\ )?
(?P<day>\d{1,2})?,?\ ?
{YEAR_REGEX}?
```
[...]

PR #330, opened 2026-08-01
fix(regexes): parse law edition/supplement parentheticals and widen publisher token

### LINKED ISSUE #325: Law-citation publisher/edition parentheticals are parsed too narrowly (publisher names, "ed."/"Supp." editions)
Parent: freelawproject/eyecite#322

### Summary

The publisher/edition/date parenthetical on a law citation is parsed by a single regex (`POST_LAW_CITATION_REGEX` in `regexes.py`), and it is too narrow. It handles a single-word publisher + plain date (`(West 2020)`) but drops or misfiles the rest. All of it is edition/date metadata a downstream resolver needs to know *which* version of the text was cited (cf. freelawproject/foresight#32).

The regex is shared by all law citations, but the failures split into a **federal** family — this issue's deliverable, per #322 — and a **state** family with the same root cause.

> Supersedes the original scope, which covered only the official `(Supp. II 1997)` form.

### Federal: U.S. Code (eyecite 2.7.8)

| Text | Expected | eyecite result |
|---|---|---|
| `18 U.S.C. § 1 (West 2020)` | pub=West, year=2020 | ✅ correct |
| `18 U.S.C. § 1 (West Jan. 1, 2012)` | pub+month+day+year | ✅ correct |
| `18 U.S.C. § 1 (2018 ed.)` | year=2018 | ❌ everything dropped |
| `8 U.S.C. § 1101 (Supp. II 1997)` | year=1997 (+ supplement) | ❌ dumped to `parenthetical` |
| `8 U.S.C. § 1101 (1994 ed. & Supp. IV 1998)` | edition + years | ❌ everything dropped |
| `18 U.S.C.S. § 1101 (LexisNexis 2018)` | pub=LexisNexis, year=2018 | ❌ dumped to `parenthetical` |
| `18 U.S.C. § 1 (West 1994 & Supp. 1997)` | pub=West + years | ❌ dumped to `parenthetical` |

Two failure modes:

1. **Official edition vocabulary** — `ed.`, `Supp.` + Roman numeral, and combined `X ed. & Supp. Y`. This is how the U.S. Code marks temporal versions between its six-year main editions; the year is lost entirely.
2. **Publisher names with internal capitals** — `LexisNexis`, publisher of the U.S.C.S. The publisher token matches only `[A-Z][a-z]+`, stops at the second capital, and the whole parenthetical falls through to `PARENTHETICAL_REGEX`.

### Minimal reproduction

```python

for t in [
    "18 U.S.C. § 1 (2018 ed.)",                    # 'ed.' -> year lost
    "8 U.S.C. § 1101 (Supp. II 1997)",             # official Supplement -> year lost
    "8 U.S.C. § 1101 (1994 ed. & Supp. IV 1998)",  # official long form -> all dropped
    "18 U.S.C.S. § 1101 (LexisNexis 2018)",        # internal capital -> publisher lost
    "18 U.S.C. § 1 (West 2020)",                   # control -> works
]:
    for c in get_citations(t):
        m = c.metadata
        print(f"{t!r:46} -> publisher={m.publisher!r} year={m.year!r} "
              f"parenthetical={m.parenthetical!r}")
```

### State codes: same root cause, broader impact

The same publisher-token bug hits the standard publishers of entire state codes. Not part of the federal deliverable, but a fix to the shared regex will likely cover these and must not regress them:

- `N.Y. Penal Law § 120.00 (McKinney 2019)` → dumped to `parenthetical`
- `Tex. Penal Code Ann. § 22.01 (Vernon's 2003)` (apostrophe) → dumped to `parenthetical`
- `Conn. Gen. Stat. Ann. § 52-570b (West Group 2000)` (multi-word) → everything dropped

### Root cause

In `POST_LAW_CITATION_REGEX`:
- `(?P<publisher>[A-Z][a-z]+\.?(?:\ Supp\.)?)` matches exactly one capitalized word — no internal capitals, apostrophes, or multi-word names.
- There is no alternative for an edition marker (`ed.`), a leading `Supp.` + Roman numeral, or the combined `X ed. & Supp. Y` form.

### Suggested direction

- Add an edition alternative capturing `ed.`, `Supp.` + optional Roman numeral, and combined `X ed. & Supp. Y`, extracting `year` (and ideally the edition/supplement designation as a field).
- Broaden the `publisher` token to allow internal capitals, apostrophes, and multi-word names (or match against a small known-publisher set).

### Real-world usage (CourtListener)

Single-year, punctuation-insensitive phrase counts (undercount the true multi-year footprint):

- Official U.S.C. supplements: `(Supp. IV 1998)` — 483 opinions · `(Supp. V 1999)` — 345 · `(Supp. II 1997)` — 26 — e.g. [United States v. Mangahas](https://www.courtlistener.com/opinion/4465564/united-states-v-mangahas/), [Pikulin v. United States](https://www.courtlistener.com/opinion/7300402/pikulin-v-united-states/)
- [`(LexisNexis 2018)`](https://www.courtlistener.com/?q=%22(LexisNexis%202018)%22&type=o) — 329
- State: [`(McKinney 2019)`](https://www.courtlistener.com/?q=%22(McKinney%202019)%22&type=o) — 27 · [`(West Group 2000)`](https://www.courtlistener.com/?q=%22(West%20Group%202000)%22&type=o) — 38 · [Vernon's](https://www.courtlistener.com/?q=%22(Vernon's%202003)%22&type=o) counts unreliable (surname collision) but the form is pervasive in Tex. statutory cites

### Scope / relation

- Metadata capture/normalization only; interpreting the edition to fetch a specific historical text is downstream (CourtListener), per freelawproject/foresight#32.
- **Not** included here — tracked separately: a bare date parenthetical `(May 2, 1999)` mis-captures the month as `publisher` (a correctness regression, distinct fix).
- Frequently co-occurs with the `note` form (freelawproject/eyecite#323).

jorgel-mendes:
## What

Fixes #325. `POST_LAW_CITATION_REGEX` dropped or misfiled most publisher/edition
parentheticals on law citations:

```python
get_citations("18 U.S.C.S. § 1101 (LexisNexis 2018)")[0].metadata
# publisher=None  year=None  parenthetical='LexisNexis 2018'   -> misfiled as prose

get_citations("18 U.S.C. § 1 (2018 ed.)")[0].metadata
# publisher=None  year=None  parenthetical=None                -> lost entirely
```

## Root cause

The publisher/date parenthetical is handled by a single group in
`POST_LAW_CITATION_REGEX`, and every component inside that group is optional.
So when it cannot match the text it does not fail — it matches *empty*, and the
trailing `PARENTHETICAL_REGEX` (`.*`) then consumes the whole parenthetical as
prose commentary. Nothing raises; the metadata is just silently wrong.

Two things it cannot match:

1. **Edition vocabulary.** There is no branch for `ed.`, for a leading `Supp.`
   + roman numeral, or for the combined `X ed. & Supp. Y` form.
2. **Publisher names that are not one plain capitalised word.** `[A-Z][a-z]+`
   is a capital followed by lowercase letters only, so it stops at the second
   capital in `LexisNexis` and at the apostrophe in `Vernon's`.

The `ed.` forms then hit a second step. `process_parenthetical()` discards any
parenthetical beginning with a four-digit year, treating it as a bare year
parenthetical — so `(2018 ed.)` and `(1994 ed. & Supp. IV 1998)` were not
merely misfiled into `parenthetical`, they were dropped entirely.

## Fix

`eyecite/regexes.py` only — three optional branches and one widened class:

- Leading supplement before the year, for `(Supp. II 1997)`.
- Main edition marker `(?:\ ed\.)?` after it, for `(2018 ed.)`.
- Trailing supplement, for `(1994 ed. & Supp. IV 1998)` and
  `(West 1994 & Supp. 1997)`.
- Publisher `[A-Z][a-z]+` → `[A-Z][A-Za-z'’]+` — internal capitals and
  apostrophes, two-character minimum kept. The roman numeral is factored out as
  `LAW_SUPPLEMENT_NUMBER_REGEX`.

Now parsed:

| parenthetical | extracted |
|---|---|
| `(2018 ed.)` | `year='2018'` |
| `(Supp. II 1997)` | `year='1997'` |
| `(1994 ed. & Supp. IV 1998)` | `year='1994'` |
| `(West 1994 & Supp. 1997)` | `publisher='West'`, `year='1994'` |
| `(LexisNexis 2018)` | `publisher='LexisNexis'`, `year='2018'` |
| `(McKinney 2019)` | `publisher='McKinney'`, `year='2019'` |
| `(Vernon's 2003)` | `publisher="Vernon's"`, `year='2003'` |

The last two are the state cases listed as "must not regress" — they fall out
of the shared publisher token.

## Scope

- `year` holds the **main edition** year; the supplement year is matched but
  not captured, since there is one `year` field.
- **No new metadata field** for the edition designation (the issue's
  "ideally") — it would also need `corrected_citation_full()`, which rebuilds
  the cite from `(publisher, month, day, year)`. Happy to add as a follow-up.
- **#326 untouched** — `(May 2, 1999)` still yields `publisher='May'`; fixing it
  means reordering publisher vs month. Verified identical before/after.
- **Multi-word publishers still unmatched.** `(West Group 2000)` falls through
  as before: a trailing word would let the publisher swallow the month in
  `(West Jan. 1, 2012)`.
- **`(Supp. V 1999)` still yields no year**, and this one cannot be fixed in
  the regex. `'v'` is in `STOP_WORDS`, so the tokenizer emits a
  `StopWordToken` inside the parenthetical, and `add_law_metadata` scans with
  `match_on_tokens(..., strings_only=True)`, which stops at the first
  non-string token — the year is never in the text the regex receives. The
  regex does match the form in isolation:

  ```python
  re.search(rf"^(?:{POST_LAW_CITATION_REGEX})", " (Supp. V 1999)", flags=re.X)
  # -> {'year': '1999'}
  ```

  `Supp. I/II/III/IV/VI/X` are unaffected — `V` is the only supplement number
  that collides with a stop word. Worth flagging since it is the second most
  common form in the issue's counts (345 opinions). A fix belongs in
  tokenization, so I have left it out here; happy to open a separate issue.

## Tests

- Six new pairs in `test_find_law_citations`, across all three tokenizers. No
  `McKinney` case — `N.Y. <subject> Law` extracts different groups per
  tokenizer on `main`, unrelated to this regex; `LexisNexis` covers the
  internal-capital path.
- Re-checked `(West 2009)`, `(West Jan. 1, 2012)`, `(Lexis Supp. 2010)`,
  `(2007)`, `(2005-06)`, and `(repealed)` still reaching `parenthetical`.
- Full suite: 54 tests pass. `ruff` check and format clean.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,15 @@ Changes:
 -
 
 Fixes:
--
+- Parse official U.S. Code edition and supplement parentheticals, and broaden
+  the law-citation publisher token. `POST_LAW_CITATION_REGEX` now extracts
+  `year` from `(2018 ed.)`, `(Supp. II 1997)`, `(1994 ed. & Supp. IV 1998)` and
+  `(West 1994 & Supp. 1997)`, where previously the whole parenthetical was
+  either misfiled as a `parenthetical` or dropped entirely. The publisher token
+  now accepts internal capitals and apostrophes, so `(LexisNexis 2018)` and
+  `(Vernon's 2003)` are captured as publishers. Where a citation carries both a
+  main edition and a supplement year, `year` holds the main edition year.
+  Multi-word publishers such as `(West Group 2000)` are still not matched. #325
 
 ## Current
 
--- a/eyecite/regexes.py
+++ b/eyecite/regexes.py
@@ -357,6 +357,10 @@ POST_SHORT_CITATION_REGEX = rf"""
 """
 
 
+# Roman numeral numbering an official U.S. Code supplement, like "Supp. II".
+# Optional because the shorthand "Supp. 1997" also occurs.
+LAW_SUPPLEMENT_NUMBER_REGEX = r"(?:[IVXL]+\ )?"
+
 # Post law citation regex:
 # statutory and regulatory cites may have publishers and dates after them, like
 #  (West), (West 1999), (Lexis Jun. 2018), (1999), or (May 2, 1999),
@@ -365,9 +369,11 @@ POST_LAW_CITATION_REGEX = rf"""
     {LAW_PIN_CITE_REGEX}?
     \ ?
     (?:\(
-        # Consol., McKinney, Deering, West, LexisNexis, etc.
+        # Consol., McKinney, Deering, West, LexisNexis, Vernon's, etc.
+        # Internal capitals and apostrophes allowed; single word only --
+        # a trailing word would swallow the month in "(West Jan. 1, 2012)".
         (?P<publisher>
-            [A-Z][a-z]+\.?
+            [A-Z][A-Za-z'’]+\.?
             (?:\ Supp\.)?
         )?
         \ ?
@@ -375,8 +381,15 @@ POST_LAW_CITATION_REGEX = rf"""
         (?:{MONTH_REGEX}\ )?
         # day
         (?P<day>\d{{1,2}})?,?\ ?
+        # leading supplement, like "Supp. II 1997"
+        (?:Supp\.\ {LAW_SUPPLEMENT_NUMBER_REGEX})?
         # four-digit year
         {YEAR_REGEX}?
+        # main edition marker, like "2018 ed."
+        (?:\ ed\.)?
+        # trailing supplement, like "& Supp. IV 1998". Its year is matched
+        # but not captured; `year` holds the main edition year.
+        (?:\ &\ Supp\.\ {LAW_SUPPLEMENT_NUMBER_REGEX}\d{{4}})?
     \))?
     \ ?
     # parenthetical
--- a/eyecite/regexes.py
+++ b/eyecite/regexes.py
@@ -358,7 +358,7 @@ POST_SHORT_CITATION_REGEX = rf"""
 
 
 # Roman numeral numbering an official U.S. Code supplement, like "Supp. II".
-# Optional because the shorthand "Supp. 1997" also occurs.
+# Optional for the trailing form, where "& Supp. 1997" occurs.
 LAW_SUPPLEMENT_NUMBER_REGEX = r"(?:[IVXL]+\ )?"
 
 # Post law citation regex:
--- a/tests/test_FindTest.py
+++ b/tests/test_FindTest.py
@@ -1015,6 +1015,47 @@ class FindTest(TestCase):
              [law_citation('Mass. Gen. Laws ch. 1, §§ 2-3',
                            reporter='Mass. Gen. Laws',
                            groups={'chapter': '1', 'section': '2-3'})]),
+            # Official main edition marker, like "(2018 ed.)" #325
+            ('18 U.S.C. § 1 (2018 ed.)',
+             [law_citation('18 U.S.C. § 1 (2018 ed.)',
+                           reporter='U.S.C.',
+                           groups={'title': '18', 'section': '1'},
+                           year=2018)]),
+            # Official supplement preceding the year #325
+            ('8 U.S.C. § 1101 (Supp. IV 1998)',
+             [law_citation('8 U.S.C. § 1101 (Supp. IV 1998)',
+                           reporter='U.S.C.',
+                           groups={'title': '8', 'section': '1101'},
+                           year=1998)]),
+            # Combined main edition and supplement; `year` is the main
+            # edition year #325
+            ('8 U.S.C. § 1101 (1994 ed. & Supp. IV 1998)',
+             [law_citation('8 U.S.C. § 1101 (1994 ed. & Supp. IV 1998)',
+                           reporter='U.S.C.',
+                           groups={'title': '8', 'section': '1101'},
+                           year=1994)]),
+            # Publisher plus trailing supplement #325
+            ('18 U.S.C. § 1 (West 1994 & Supp. 1997)',
+             [law_citation('18 U.S.C. § 1 (West 1994 & Supp. 1997)',
+                           reporter='U.S.C.',
+                           metadata={'publisher': 'West'},
+                           groups={'title': '18', 'section': '1'},
+                           year=1994)]),
+            # Publisher with an internal capital #325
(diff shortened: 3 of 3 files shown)


mlissner:
To Bill for visibility!

mlissner:
@claude review

flooie:
@jorgel-mendes 

A few things I'd like to work through before this lands.

Supp. V still isn't working, and I'd question the premise.

StopWordToken exists to find the v. in a case name — every consumer of it in helpers.py is case-name machinery. It has no meaning for a law citation at all. So the scan isn't stopping for a good reason; it's stopping because a generic flag doesn't know the difference. Take another look at the call you already identified and I think you'll find a fix that stays in helpers.py and never touches the tokenizer.

The bigger issue: we're dropping the edition data, and it won't work downstream.

Worth spelling out why this matters, because I think it reframes the whole ticket. The U.S. Code is published in main editions every six years — 1994, 2000, 2006, 2012, 2018.  Between
  those, the supplements (Supp. I through V) carry the intervening changes. Critically, a supplement contains only the changes, not the full text. So when a judge writes:

8 U.S.C. § 1101 (1994 ed. & Supp. IV 1998)

they are citing two books: the 1994 main edition for the base text, and Supplement IV for what changed by 1998. You need both to reconstruct the provision they actually read. Right
  now we return year=1994 and discard the rest which points a downstream resolver at the 1994 text the version before the amendment the judge is relying on. That's not  incomplete metadata, it's metadata that resolves to the wrong law.

Which is why I don't think a single year is the right model.  And I think the original ticket is pointing to a necessary change to the metadata fields in the form of adding [edition].

Im highlighting the rarer example of two editions, but I think the logic holds for `1994 ed` as an example all the same.

flooie (changes requested):
@jorgel-mendes 

This is a nice first attempt, but I think it needs some tweaks and a little more work to get it over the line.  I laid out most of my commentary in a comment.

PR #333, opened 2026-08-07
fix(ci): run the benchmark on fork and Dependabot PRs

### LINKED ISSUE #332: `benchmark.yml` fails on every fork and Dependabot PR
The benchmark dies on its first step for any PR not pushed by someone with write access, so outside contributions never get benchmarked.

## Impact

**66 of the last 81 `pull_request` runs failed (81%)**, across 19 branches, back to at least 2025-06-24. All 15 successes were same-repo branches: **no fork PR has ever produced a benchmark.** Currently affected: #328, #331, #330, #308.

## Root cause

Every failure is step 6 of 16, `Add or Update comment on PR that Test is running` ([example](https://github.com/freelawproject/eyecite/actions/runs/30675305060)):

```
##[error]Resource not accessible by integration - .../issues/comments#create-an-issue-comment
```

`GITHUB_TOKEN` is scoped to the *triggering event*: read-only for fork PRs and for Dependabot, writable for same-repo pushes by humans. The sticky comment needs `issues: write`, so it 403s — and because it's the *first* step, the benchmark never starts.

Neither workaround exists today: re-running replays the original event (confirmed on #328 — 200 from the rerun API, same failure 15s later), and there's no `workflow_dispatch` trigger (`gh workflow run` → `HTTP 422`).

## Fix: split the workflow in two

One job currently does two things needing opposite permissions: it **runs the contributor's code** (must be unprivileged, or a fork can take our token and secrets) and it **comments and pushes results** (needs write). They can't share a job.

`workflow_run` is a trigger that fires when a named workflow completes. It always runs the copy of the file on the default branch, with a normal write token and secrets — regardless of what triggered the original run. That gives us the split:

| | `benchmark.yml` (existing) | `benchmark-comment.yml` (new) |
|---|---|---|
| Trigger | `pull_request` | `workflow_run`, when the first finishes |
| Runs the PR's code | yes | **never** |
| Token | read-only, no secrets | write + secrets |
| Job | run both benchmark passes, upload `results/` as a build artifact | download that artifact, post the comment, push to `artifacts` |

The untrusted half produces files; the trusted half publishes them without executing anything from the fork. That second workflow must never check out or run PR code — that's the property the whole design rests on.

Two things that will bite:

- `github.event.workflow_run.pull_requests` is **empty for fork PRs**, so the PR number has to travel inside the artifact as a plain file.
- `workflow_run` only ever runs the **default branch's** copy, so it can't be tested from a PR; expect a shakedown PR after merge.

<details>
<summary>Why not <code>pull_request_target</code></summary>

It grants a write token and secrets, and this workflow deliberately executes the PR's code (`cp "$RUNNER_TEMP/benchmark.py" benchmark/benchmark.py` then runs it, plus `uv sync --frozen` on the PR's lockfile). That's arbitrary code execution with `FREELAWBOT_TOKEN` in the environment.

</details>

## Maintainer gate: `run-benchmark` label

Maintainers keep automatic runs; a label lets them opt a fork PR in.

```yaml
on:
  pull_request:
    types: [opened, synchronize, reopened, labeled]

jobs:
  benchmark:
    if: >-
      github.event.pull_request.head.repo.full_name == github.repository
      || github.event.pull_request.author_association == 'MEMBER'
      || github.event.pull_request.author_association == 'OWNER'
      || contains(github.event.pull_request.labels.*.name, 'run-benchmark')
```

Nobody on the team changes how they work: across the last 45 PRs, every maintainer PR still runs automatically and only fork PRs need the label.

**Gate on the head repo, not on `author_association`.** Checking `author_association in (OWNER, MEMBER, COLLABORATOR)` would block **all six Dependabot PRs** (#327, #314, #313, #312, #311, #310) — Dependabot reports `CONTRIBUTOR`, since the field describes contribution history, not access. `COLLABORATOR` appears zero times in those 45 PRs. "Branch lives in our repo" implies write access and matches the boundary GitHub already uses for the token.

Wrinkles:

- Specifying `types:` overrides the default, so `reopened` must be listed or it's silently lost.
- `labeled` fires for *any* label; adding `bug` re-triggers. Harmless with `concurrency`, or guard with `github.event.action != 'labeled' || github.event.label.name == 'run-benchmark'`.
- Once labeled, later pushes re-run via `synchronize`. For per-push approval, have the `workflow_run` job strip the label after each run.

The label decides *whether* it runs, not *what permissions* it gets — a labeled fork PR still gets a read-only token, so the split above is still required.

## Related, out of scope here

- **Fork-PR approval setting is off**, worth its own admin-only issue: #328, #330, #331 are all first-time contributors with 0 merged PRs, yet their runs started within ~9s of the push, so **Settings → Actions → "Fork pull request workflows from outside collaborators"** isn't even at GitHub's default. This matters more for `tests.yml`, which runs arbitrary fork code on every PR with no gate at all — gating only the benchmark is a cost control, not a security control.
- **`benchmark.yml` has no `concurrency`, `timeout-minutes`, or `permissions:` block.** `tests.yml:43-45` already has the pattern; without it force-pushes spawn unbounded parallel runs (#331 racked up 5).
- **`dmnemec/copy_file_to_another_repo_action@main`** is pinned to a moving ref and handed `FREELAWBOT_TOKEN` — worth pinning to a SHA as it moves into the privileged workflow.
- #212 (benchmark readability/stats), #158 (same permissions breakage, closed 2024 without a structural fix).

grossir:
The benchmark job both ran the PR's code and commented on the PR, so it needed a writable token in a context where GitHub only grants a read-only one. It died at its first step for every fork and Dependabot PR -- 66 of the last 81 runs -- and no fork PR has ever produced a benchmark.

Split the privileged half into publish-benchmark.yml, which runs from the default branch on workflow_run and never checks out PR code. benchmark.yml uploads its results as an artifact instead, and holds only `contents: read`.

Fork PRs are gated behind a `run-benchmark` label; anything pushed to this repo (every maintainer, plus Dependabot) still runs automatically. The gate keys on the head repo rather than author_association, which reports contribution history rather than access -- Dependabot reports CONTRIBUTOR.

Also adds concurrency groups so a new push supersedes a running benchmark, 15 minute job timeouts in place of the 6 hour default, and pins copy_file_to_another_repo_action to a commit SHA instead of a mutable ref.

Fixes #332

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -36,8 +36,19 @@ jobs:
     # `author_association` reports contribution history rather than access --
     # Dependabot reports CONTRIBUTOR -- so the head-repo check is what keeps
     # everyone with write access running automatically. #332
+    #
+    # The `labeled` clause keeps an unrelated label from re-running the
+    # benchmark: `labeled` fires for every label, so on a PR that already
+    # passes the gate below, adding `bug` triggered a full re-run -- and with
+    # `cancel-in-progress` that cancels the in-flight benchmark rather than
+    # merely duplicating it. It has to be `action != 'labeled' || ...` rather
+    # than a plain equality, because `github.event.label` is absent on
+    # opened/synchronize/reopened; comparing it there would be false and the
+    # benchmark would only ever run on labeling.
     if: >-
       github.event_name == 'pull_request'
+      && (github.event.action != 'labeled'
+      || github.event.label.name == 'run-benchmark')
       && (github.event.pull_request.head.repo.full_name == github.repository
       || github.event.pull_request.author_association == 'MEMBER'
       || github.event.pull_request.author_association == 'OWNER'
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,17 +1,51 @@
 name: Benchmark Pull Request
 on:
   pull_request:
+    # `labeled` lets a maintainer opt a fork PR in (see the `if:` below). Once
+    # `types:` is set the defaults no longer apply, so they must be listed.
+    types: [opened, synchronize, reopened, labeled]
   repository_dispatch:
     types: [ reporters-db-pr ]
 
 env:
   main: "$(/usr/bin/git log -1 --format='%H')"
 
+# A new push supersedes any benchmark still running for the same PR, so only
+# the newest commit is measured. Keyed on the PR number rather than
+# `github.head_ref` (as tests.yml uses) because two forks can share a branch
+# name -- `patch-1` is GitHub's default for web edits -- and would otherwise
+# cancel each other. `repository_dispatch` runs have no PR number and fall
+# back to a unique run id, so they never cancel one another.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+# This workflow runs the PR's own code, so it must never hold a writable
+# token: for fork PRs GitHub already forces one that is read-only, and we
+# declare it here so the same is true for internal branches. Commenting and
+# pushing results happen in publish-benchmark.yml, which never runs PR code.
+permissions:
+  contents: read
+
 jobs:
   benchmark:
-    name: PR comment
-    if: github.event_name == 'pull_request'
+    name: Benchmark
+    # Run automatically for anything pushed to this repo (all maintainers, plus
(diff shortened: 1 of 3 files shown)


quevon24 (approved):
LGTM. Splitting the privileged half into its own workflow is the right fix, and I checked the action pin and versions. One non-blocker: `labeled` fires for any label, so adding `bug` to a PR re-runs the benchmark, worth adding the `github.event.label.name == 'run-benchmark'` guard from #332

Issue #321, opened 2026-07-08
Metadata mis-association across neighboring citations: plaintiff/defendant and year attach to the wrong FullCaseCitation


vappsusa:
**eyecite version:** 2.7.8 · Python 3.14 · AhocorasickTokenizer

When several case citations appear near each other, `FullCaseCitation.metadata` can pick up the **previous** citation's caption parties or year parenthetical instead of its own.

### Repro 1 — party name from the previous sentence

```python
from eyecite import get_citations
from eyecite.tokenizers import AhocorasickTokenizer

text = (
    'The court in Smith v. Jones, 347 U.S. 999 (1954), held that '
    '"schools must serve pizza." Smith v. Jones, 347 U.S. 999 (1954). '
    'Compare Brown v. Board of Education, 347 U.S. 483 (1954).'
)
for c in get_citations(text, tokenizer=AhocorasickTokenizer()):
    print(c.corrected_citation(), repr(c.metadata.plaintiff), repr(c.metadata.defendant))
```

Output:
```
347 U.S. 999 'Smith' 'Jones'
347 U.S. 999 'Smith' 'Jones'
347 U.S. 483 '' 'Jones'          # <- Brown's cite, defendant from the prior sentence
```
The third citation is captioned "Brown v. Board of Education" in the text, but its metadata reports `defendant='Jones'` (and `plaintiff=''`) from the preceding case. `full_span()` for that citation also stretches back across the sentence boundary.

### Repro 2 — year from the previous citation's parenthetical

```python
text = "Miranda v. Arizona, 347 U.S. 483 (1990). NLRB v. IBEW, 346 U.S. 464 (1953)."
for c in get_citations(text, tokenizer=AhocorasickTokenizer()):
    print(c.corrected_citation(), c.metadata.year)
```

Output:
```
347 U.S. 483 1990
346 U.S. 464 1990        # <- text says (1953) immediately after this cite
```
The second citation's own `(1953)` parenthetical is ignored and it inherits `1990` from the first citation.

### Why it matters
We use party/year metadata to cross-check citations in AI-generated legal text (wrong-case-name and wrong-year detection). Mis-associated metadata makes a correctly cited case look mis-attributed — we currently work around it by re-anchoring both fields against the text window adjacent to each citation's `span()`, but a fix at the extraction layer would benefit everyone consuming `metadata`.

Happy to provide more cases or test against a branch. Thanks for eyecite — it's load-bearing for us.

PR #339, opened 2026-08-13
fix(tokenizers): relax whitespace at full-cite template boundaries

### LINKED ISSUE #338: Glued citations (`846F.2d746`): relax whitespace at the full-cite template boundaries -- completes #305
## Summary

I'm building citation-verification tooling on top of eyecite (checking
legal filings for fabricated/unverifiable citations) and hit a class of
real-world citations that extract as nothing.

eyecite requires exactly one literal space at the volume-reporter and
reporter-page boundaries, so a "glued" citation -- `846F.2d746`,
`410U.S.113` -- produces no match at all. #305/#307 relaxed whitespace
*within* reporter abbreviations; the two template boundaries are the
remaining rigid seams. This proposal extends the same `\s*` relaxation to
them.

The downstream failure mode is a false negative: a verifier reports
"0 problems found" on a document whose citations it never saw, and glued
citations are missing from any citation graph built with eyecite.

## Prevalence

From a scan of the full CourtListener bulk opinion export (10,798,347
opinions): **7,740 opinions** contain glued citations (10,488 matches,
4,683 distinct strings). Two provenances: OCR-introduced, and born-digital
(a 2016 S.D. Tex. filing carries `42U.S.C. §1981` in the PDF's own text
layer, next to correctly-spaced `42 U.S.C. § 2000e`). Glue also
propagates: `846F.2d746` appears in 68 distinct RECAP filings via
copy-paste. I can contribute document IDs + context windows as regression
fixtures.

## Root cause

`_relax_ws` (#307) applies only to edition names; the template keeps
literal spaces at the boundaries:

```python
raw_regex_variables["full_cite"][""] = "$volume $reporter,? $page"   # tokenizers.py L138
```

## Proposed change

```python
raw_regex_variables["full_cite"][""] = r"$volume\s*$reporter,?\s*$page"
```

plus trimming the trailing `\s*` in `_relax_ws` output. The trim is
load-bearing: without the mandatory boundary space, the greedy `\s*`
after a reporter's final period captures the following space into the
reporter group (`reporter='U.S. '`), breaking every consumer of the
groups (~190 test failures). With the trim, the test-suite failure set is
byte-identical to master on my environment.

| input | master | patched |
|---|---|---|
| `846F.2d746` | (none) | FullCaseCitation |
| `410 U.S. 113` | FullCaseCitation | unchanged |
| `799 N. Y. S. 2d 795` (#305 style) | FullCaseCitation | unchanged |
| `7F., No. 106, Sec. 5` (address) | (none) | (none) |

## Measured gains / costs

Replaying master vs patched over the stored context windows of all 7,740
hit documents (11,748 windows):

- **Gains:** 7,737 new citations across 4,141 documents; 87% carry
  plausible volumes (`846F.2d746`, `443U.S.307`, `550U.S. 544`, short
  cites like `757F.2d at 1241`).
- **Cost:** 13% have implausible volumes -- digit fusion (`544,1127S. Ct.
  1955`; true cite `127 S. Ct. 1955`): a digit glued ahead of the volume
  is absorbed into it, and no tokenizer rule can recover the split. All
  of these occur where master extracts nothing, and an implausible volume
  fails any downstream lookup -- a visible wrong guess replacing an
  invisible miss. Volume-plausibility gating would live in validation; I'd
  scope it as a follow-up.
- **Losses:** effectively zero -- 8 overlap swaps (recovered cite retires
  a redundant `SupraCitation`), 4 garbage `UnknownCitation` blobs replaced
  by real parses, 1 double-glue pathology (`347U.S. 483,349 U.S. 294`:
  first cite recovered, its pin-cite parser swallows the second's volume).

Out of scope: statute glue (`42U.S.C.§1983`) rides the laws templates and
still degrades to `UnknownCitation` -- separate issue, likely reporters-db.

I have the patch and replay tooling ready and am happy to open a PR.

jaysoma:
## Fixes

Fixes: #338

## Summary

Glued citations (`846F.2d746`) currently extract as nothing: the full-cite
template requires exactly one literal space at the volume-reporter and
reporter-page boundaries. This extends #305's whitespace relaxation to those
two boundaries:

- `full_cite` template becomes `r"$volume\s*$reporter,?\s*$page"`
- `_relax_ws` now trims its trailing `\s*` so the reporter group cannot
  capture boundary whitespace (`reporter='U.S. '`) once the mandatory
  template space is gone.

Measured on the 7,740 glued-citation opinions from the bulk scan (details in
#338): 7,737 recovered citations across 4,141 documents, no losses of
previously-extracted citations, existing test suite unchanged. New regression
tests cover glued full/short cites, composition with #305, group cleanliness,
and a non-citation negative.

## AI Disclosure

- [ ] No AI tools were used to create the content of this PR.
- [x] Parts of this PR were created with the help of an AI tool, and I have carefully reviewed all of its content and take full responsibility for it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,10 @@ Fixes:
 - Fix 2 `TypeError`s raised in `get_citations(markup_text=...)` when
   `clean_steps` was omitted or lacked `"html"`. The documented fallback that
   prepends the `html` step was both unreachable and broken.
+- Match "glued" citations (`846F.2d746`) by relaxing whitespace at the
+  full-cite template's volume-reporter and reporter-page boundaries,
+  completing the #305 relaxation; `_relax_ws` now trims its trailing `\s*`
+  so the reporter group never captures boundary whitespace. #338
 
 ## Current
 
@@ -57,7 +61,7 @@ Features:
 
 Changes:
 - Move dependency management to uv.
-  This shouldn’t have any visible impact to users, except from a few small metadata changes.
+  This shouldn't have any visible impact to users, except from a few small metadata changes.
 
 Fixes:
 - Fixes rendering of AhocorasickTokenizer parameter definition in API docs #279
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,13 @@ Fixes:
   full-cite template's volume-reporter and reporter-page boundaries,
   completing the #305 relaxation; `_relax_ws` now trims its trailing `\s*`
   so the reporter group never captures boundary whitespace. #338
+- Match both of two adjacent citations that share a single separator
+  character (`347 U.S. 483,349 U.S. 294` silently lost the second
+  citation): `TokenExtractor.get_matches` resumes scanning at the end of
+  the token content, so a separator consumed as one match's trailing
+  boundary stays available as the next match's leading boundary. This
+  aligns the pure-Python tokenizers with `HyperscanTokenizer`, which
+  reports overlapping matches and never had this blind spot. #338
 
 ## Current
 
--- a/eyecite/models.py
+++ b/eyecite/models.py
@@ -850,8 +850,25 @@ class TokenExtractor:
     strings: list = field(default_factory=list)
 
     def get_matches(self, text):
-        """Return match objects for all matches in text."""
-        return self.compiled_regex.finditer(text)
+        """Return match objects for all matches in text.
+
+        Successive searches resume from the end of the token content
+        (group 1) rather than the end of the full match: boundary
+        wrappers like nonalphanum_boundaries_re consume one character
+        on each side, so under finditer semantics a separator consumed
+        as one match's trailing boundary is unavailable as the next
+        match's leading boundary, and the second of two adjacent
+        citations is silently skipped (e.g. the parallel cite
+        "347 U.S. 483,349 U.S. 294" loses "349 U.S. 294"). Hyperscan
+        reports overlapping matches, so HyperscanTokenizer does not
+        share this blind spot; resuming at the content boundary keeps
+        the two tokenizers consistent.
+        """
+        pos = 0
+        while m := self.compiled_regex.search(text, pos):
+            yield m
+            end = m.end(1) if m.re.groups and m.end(1) != -1 else m.end()
+            pos = end if end > pos else pos + 1
 
     def get_token(self, m, offset=0) -> Token:
         """For a given match object, return a Token."""
--- a/eyecite/tokenizers.py
+++ b/eyecite/tokenizers.py
@@ -117,6 +117,12 @@ def _relax_ws(escaped: str) -> str:
     # (`re.escape` escapes a space as "\ "; match it with or without the
     # backslash so this is robust across Python versions.)
     escaped = re.sub(r"(?:\\?\ )+", r"\\s*", escaped)
+    # Drop a trailing optional-whitespace so the abbreviation pattern never
+    # captures whitespace past its final token: with flexible whitespace at
+    # the template boundaries there is no following literal space to force
+    # the greedy \s* to backtrack, so e.g. `U\.S\.\s*` would capture
+    # "U.S. " (trailing space) into the reporter group.
+    escaped = re.sub(r"(?:\\s\*)+$", "", escaped)
     return escaped
 
 
@@ -135,7 +141,7 @@ def _populate_reporter_extractors():
 
     # Set up regex replacement variables from reporters-db
     raw_regex_variables = deepcopy(RAW_REGEX_VARIABLES)
-    raw_regex_variables["full_cite"][""] = "$volume $reporter,? $page"
+    raw_regex_variables["full_cite"][""] = r"$volume\s*$reporter,?\s*$page"
     raw_regex_variables["page"][""] = rf"(?P<page>{PAGE_NUMBER_REGEX})"
     regex_variables = process_variables(raw_regex_variables)
 
--- a/tests/test_FindTest.py
+++ b/tests/test_FindTest.py
@@ -954,6 +954,36 @@ class FindTest(TestCase):
         # fmt: on
         self.run_test_pairs(test_pairs, "Glued citations")
 
+    def test_adjacent_citations_share_boundary(self):
+        """Do two citations separated by a single character both match?
+
+        The nonalphanum boundary wrappers consume one character on each
+        side of a match, so under finditer semantics a separator consumed
+        as one citation's trailing boundary was unavailable as the next
+        citation's leading boundary, and the second citation was silently
+        skipped. Parallel cites hit this whenever the separator is a bare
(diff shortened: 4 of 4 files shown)


jaysoma:
I ran this patch against real-world data to validate it beyond the unit tests:
a bulk scan of the CourtListener opinion dump (10.8M rows) found 7,740 opinions
containing glued citations, and I replayed both pristine main and this branch
over the stored context windows of every one of them (11,748 unique windows
across 7,489 documents after deduplication).

**Recovery:** 7,416 windows in 4,141 documents gained citations that main
extracts nothing from — 6,568 `FullCaseCitation`, 1,166 `ShortCaseCitation`,
3 `SupraCitation`. Known specimens (`846F.2d746`, `443U.S.307`,
`42U.S.C.§1983`) all extract with correct volume/reporter/page groups.

**Regressions:** the diff initially showed 14 windows where the branch lost a
citation main had found. On inspection:

- 9 are a bare `SupraCitation("supra,")` replaced by the short citation it
  pointed at (`Attaway, supra, 757F.2d at 1241-42` — main only saw the
  `supra`; the branch recovers `757 F.2d at 1241` and the overlap filter
  absorbs the supra, preserving the referent as `antecedent_guess`). Net
  information gain.
- 3 are garbage `UnknownCitation` blobs (e.g.
  `544(2007);28U.S.C.§1915.Dismissal…`) replaced by real citations.
- 1 was a true regression, now fixed by the second commit: in
  `347U.S. 483,349 U.S. 294` the first citation's consuming trailing
  boundary swallowed the comma, starving the second citation's leading
  boundary. Notably this swallow **exists on main for unglued text** —
  `get_citations("347 U.S. 483,349 U.S. 294")` returns only the first
  citation — and only the pure-Python tokenizers are affected;
  `HyperscanTokenizer` reports overlapping matches and never had the blind
  spot. `TokenExtractor.get_matches` now resumes scanning at the end of the
  token content rather than the end of the boundary-consuming match, which
  aligns the tokenizers and fixes both the glued and unglued cases.

**Over-capture, documented honestly:** when OCR glues a *preceding number*
onto the volume, the digits are inherently ambiguous: `…U.S. 137, 8 S. Ct.
38…` glued to `1378S. Ct. 38` extracts volume 1378. About 285 of the ~7.7k
newly matched instances (≈4%) are in this class (recognizable here by
impossible S. Ct. volumes). Two things worth noting: (1) main already
extracts volume 1378 from the properly-spaced string — eyecite has no
volume-plausibility gating anywhere, so this is pre-existing behavior that
glued input merely produces more of; (2) a fabricated-looking volume still
fails any downstream lookup, whereas main extracts nothing from these
windows at all. A volume-plausibility check against reporters-db ranges
would be a natural follow-up if there's interest.

**Also observed:** the overlap filter now sees pairings it has no rule for
(`Supra` followed by adjacent `ShortCase`, parallel `ShortCase` pairs) and
logs "Unknown overlap case" warnings; outcomes are reasonable but
order-dependent (the supra is absorbed in one processing order, kept
alongside in the other). Happy to file that separately.

Replay harness and full per-window diffs are available if useful.

jaysoma:
I ran this patch against real-world data to validate it beyond the unit tests:
a bulk scan of the CourtListener opinion dump (10.8M rows) found 7,740 opinions
containing glued citations, and I replayed both pristine main and this branch
over the stored context windows of every one of them — 11,748 windows across
7,489 documents.

**Recovery:** baseline extracts 8,750 citations; this branch extracts 16,472,
a gain of +7,722 (+88%). The gain lands in 7,405 windows across 4,138
documents — 6,564 `FullCaseCitation`, 1,157 `ShortCaseCitation`,
3 `SupraCitation`. Known specimens (`846F.2d746`, `443U.S.307`,
`42U.S.C.§1983`) all extract with correct volume/reporter/page groups.

**Regressions: zero.** Across all 11,748 windows there is no window where this
branch extracts fewer citations than main. Getting there took a second commit —
the first pass showed 14 windows that looked like losses, and they broke down as:

- 9 were a bare `SupraCitation("supra,")` replaced by the short citation it
  pointed at (`Attaway, supra, 757F.2d at 1241-42` — main only saw the
  `supra`; the branch recovers `757 F.2d at 1241` and the overlap filter
  absorbs the supra, preserving the referent as `antecedent_guess`). Net
  information gain.
- 3 were garbage `UnknownCitation` blobs (e.g.
  `544(2007);28U.S.C.§1915.Dismissal…`) replaced by real citations.
- 1 was a true regression, now fixed: in `347U.S. 483,349 U.S. 294` the first
  citation's consuming trailing boundary swallowed the comma, starving the
  second citation's leading boundary. Notably this swallow **exists on main for
  unglued text** — `get_citations("347 U.S. 483,349 U.S. 294")` returns only
  the first citation — and only the pure-Python tokenizers are affected;
  `HyperscanTokenizer` reports overlapping matches and never had the blind
  spot. `TokenExtractor.get_matches` now resumes scanning at the end of the
  token content rather than the end of the boundary-consuming match, which
  aligns the tokenizers and fixes both the glued and unglued cases.

**Over-capture, documented honestly:** when OCR glues a *preceding number* onto
the volume, the digits are inherently ambiguous: `…U.S. 137, 8 S. Ct. 38…`
glued to `1378S. Ct. 38` extracts volume 1378. About 285 of the newly matched
instances (≈4%) are in this class, recognizable by impossible S. Ct. volumes.
Two things worth noting: (1) main already extracts volume 1378 from the
properly-spaced string — eyecite has no volume-plausibility gating anywhere, so
this is pre-existing behavior that glued input merely produces more of; (2) a
fabricated-looking volume still fails any downstream lookup, whereas main
extracts nothing from these windows at all. A volume-plausibility check against
reporters-db ranges would be a natural follow-up if there's interest.

**Also observed:** the overlap filter now sees pairings it has no rule for
(`Supra` followed by adjacent `ShortCase`, parallel `ShortCase` pairs) and logs
"Unknown overlap case" warnings; outcomes are reasonable but order-dependent
(the supra is absorbed in one processing order, kept alongside in the other).
Happy to file that separately.

Replay harness and full per-window diffs are available if useful.

mightymikesapp:
Independent confirmation of the adjacent-citation half of this PR, from a different source: a pasted table of authorities (one bare citation per line) loses every second citation on 2.7.8, because the consumed boundary character on one match is the leading boundary the next one needs. This branch's `get_matches` change recovers them (newline, space, tab, and a glued comma all go from 3/5 to 5/5), and it makes AhocorasickTokenizer agree with HyperscanTokenizer, which already sees overlapping matches.

One interaction to flag: once the neighbor is recovered, it inherits the previous citation's case name, because the backward case-name scan steps over any preceding citation. On `Roe v. Wade, 410 U.S. 113\n597 U.S. 215`, `597 U.S. 215` gets plaintiff=Roe, defendant=Wade, both on this branch and when combined with #350. That's the #321 family; I've put details and a proposed approach in #350 so this PR can stay focused.

PR #341, opened 2026-08-22
fix(ci): fix GNU sed syntax in Pdoc workflow and add manual trigger

### LINKED ISSUE #316: Pdoc workflow fails on `Run sed` step (BSD vs GNU sed -i syntax)
## Problem

The `Pdoc` workflow (`.github/workflows/pdoc.yml`) fails on the `Run sed` step whenever a `v*.*.*` tag is pushed. Seen on the v2.7.7 release:

```
sed: can't read s#(AhocorasickTokenizer\().*(markup_text:)#\1),<br>\2#g: No such file or directory
##[error]Process completed with exit code 2.
```

Run: https://github.com/freelawproject/eyecite/actions/runs/28200626192/job/83538907326

This does **not** affect the actual package release — `pypi.yml` runs independently and v2.7.7 published successfully. The only impact is that the hosted API documentation site (gh-pages) is not refreshed for the failing release.

## Root cause

Line 38 uses BSD/macOS `sed` syntax:

```bash
sed -i '' -E 's#...#...#g' eyecite/find.html eyecite/index.html
```

On macOS, `-i` requires a separate backup-suffix argument (`''` = no backup). The CI runner is `ubuntu-latest`, which uses **GNU sed**, where `-i` takes no separate argument. GNU sed therefore consumes `''` as the (empty) script and treats the real `s#...#...#g` expression as a filename — hence the "can't read ..." error.

## Fix

Drop the `''` so it is valid GNU sed:

```diff
- sed -i '' -E 's#(AhocorasickTokenizer\().*(markup_text:)#\1),<br>\2#g' eyecite/find.html eyecite/index.html
+ sed -i -E 's#(AhocorasickTokenizer\().*(markup_text:)#\1),<br>\2#g' eyecite/find.html eyecite/index.html
```

## Re-deploying the 2.7.7 docs

`pdoc.yml` only triggers on a `v*.*.*` tag push, and that same trigger also fires `pypi.yml`. Re-pushing/moving the tag would re-run PyPI (fails harmlessly on duplicate upload). To avoid that, consider also adding a `workflow_dispatch:` trigger to `pdoc.yml` so the docs build can be re-run manually from the Actions tab against an existing tag.

## Acceptance criteria

- [ ] `Run sed` step succeeds on a tagged release
- [ ] (optional) `pdoc.yml` has a `workflow_dispatch` trigger for manual re-runs
- [ ] 2.7.7 API docs re-deployed to gh-pages

245678000000:
## Fixes
Fixes: #316

## Summary
The Pdoc workflow (`.github/workflows/pdoc.yml`) runs on `ubuntu-latest` (GNU sed), where `sed -i ''` causes sed to treat `''` as an empty script and the following regex argument as a target filename, resulting in `sed: can't read ...: No such file or directory`.

This PR:
1. Removes the macOS-specific `''` argument from `sed -i` so it executes properly under GNU sed.
2. Adds a `workflow_dispatch:` trigger to allow manually re-running the docs build workflow from GitHub Actions without re-pushing/moving release tags.
3. Updates `CHANGES.md`.

## AI Disclosure
- [ ] No AI tools were used to create the content of this PR.
- [x] Parts of this PR were created with the help of an AI tool, and I have carefully reviewed all of its content and take full responsibility for it.

### DIFF
--- a/.github/workflows/pdoc.yml
+++ b/.github/workflows/pdoc.yml
@@ -2,8 +2,9 @@ name: Pdoc
 
 on:
   push:
-      tags:
-        - v*.*.*
+    tags:
+      - v*.*.*
+  workflow_dispatch:
 
 jobs:
   build:
@@ -35,7 +36,7 @@ jobs:
       - name: Run sed  # Removes insanely-long parameter definition from docs
         working-directory: ./api_documentation
         run: |
-          sed -i '' -E 's#(AhocorasickTokenizer\().*(markup_text:)#\1),<br>\2#g' eyecite/find.html eyecite/index.html
+          sed -i -E 's#(AhocorasickTokenizer\().*(markup_text:)#\1),<br>\2#g' eyecite/find.html eyecite/index.html
 
       - name: Deploy 🚀
         uses: JamesIves/github-pages-deploy-action@v4
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,8 @@ Changes:
 - CI: benchmark runs are superseded when a PR is updated, time out after 15 minutes, and pin their third-party action to a commit SHA. #332
 
 Fixes:
+- CI: fix `sed` invocation syntax in the Pdoc workflow on GNU/Linux runners and
+  add `workflow_dispatch` trigger. #316
 - CI: harden the benchmark's `repository_dispatch` job against shell script
   injection by passing `client_payload` values through the environment and
   validating their shape, instead of interpolating them into `run:`. #337


.github/workflows/pdoc.yml
```diff
@@ -2,8 +2,9 @@ name: Pdoc
 
 on:
   push:
-      tags:
-        - v*.*.*
+    tags:
+      - v*.*.*
+  workflow_dispatch:
```
mlissner, line 7: What's up with this?

mlissner:
@claude review

PR #331, opened 2026-08-04
Add additional regex to resolve issue in Volunteer backlog ticket #146

linhtpham09:

Issue #146, opened 2023-03-13
Citation parser fails for statutes with letters in the section number


jmesserschmidt1:
U.S. code statutes with letters in them appear to be unrecognized. So "18 U.S.C. § 1028" and "18 U.S.C. § 1028(a)" are parsed, but "18 U.S.C. § 1028A" is not. I've tried some variations, but seems to be consistent.

mlissner:
Thanks for sending this along. I think this would be pretty easy to fix, but our code parsers aren't particularly advanced compared to our opinions parsers. 

Do you want to take a stab at it?

jmesserschmidt1:
> Thanks for sending this along. I think this would be pretty easy to fix, but our code parsers aren't particularly advanced compared to our opinions parsers.
> 
> Do you want to take a stab at it?

Sure. Not super familiar with the code, but suspect might need a variation on the law_section regex similar to the one that exists for page or volume, like [here](https://github.com/freelawproject/reporters-db/blob/22a42ce1bc8311631a7370a9efd5dc366d119c38/reporters_db/data/regexes.json#L68). This comes up with CFR cites as well (e.g., 17 CFR § 240.10b-5 is currently parsed as 17 CFR 240).  So something like `(?P<section>\\d+(?:[\\-.:]\\d+){,3})[a-zA-Z]{0,4})` and `(?P<section>\\d+(?:[\\-.:]\\d+){,3})[a-zA-Z]{0,4})`

mlissner:
I don't know that part of the code very well either, but if you want to do a PR with tests that fixes this, I think we'd probably merge it (and release a new version, if desired).

linhtpham09:
Hello, I am interested in picking up this ticket if it is still available?

mightymikesapp:
Picking this up alongside #331 and #335, which fix different halves of it. I checked out both branches and ran the same strings against main, #331, and #335:

| citation | main | #331 | #335 |
|---|---|---|---|
| 18 U.S.C. § 1028A | `§` | FullLawCitation ✓ | ShortLawCitation `§` (uppercase not matched) |
| 17 U.S.C. § 106A | `§` | FullLawCitation ✓ | ShortLawCitation `§` |
| 15 U.S.C. § 78j(b) | `§` | `§` | ShortLawCitation `§ 78j(b)` |
| 42 U.S.C. § 2000e-2 | `§` | `§` | ShortLawCitation `§ 2000e-2` |
[...]

PR #342, opened 2026-09-01
fix: extract USC citations when the section mark is omitted

### LINKED ISSUE #300: US Code citations without § symbol not extracted (e.g. `50 U.S.C. 1701`)
eyecite requires the `§` symbol to recognize U.S.C. citations. Citations of the form `50 U.S.C. 1701` return no results, while `50 U.S.C. § 1701` is correctly extracted.

## Reproduction

```python

get_citations("50 U.S.C. 1701")        # → []
get_citations("50 U.S.C. § 1701")      # → [FullLawCitation]
get_citations("18 U.S.C. 1961")        # → []
```

## Real-world example

Executive Order 13224 (published in the Federal Register at https://www.federalregister.gov/documents/2001/09/25/01-24205/blocking-property-and-prohibiting-transactions-with-persons-who-commit-threaten-to-commit-or-support) uses the no-`§` form throughout — e.g. *"...declared a national emergency with respect to the unusual and extraordinary threat...pursuant to the International Emergency Economic Powers Act (50 U.S.C. 1701 et seq.)"* — and eyecite extracts 0 USC citations from the document.

This pattern (no `§`) is common in Federal Register documents, including Executive Orders. A pre-processing normalization of `U.S.C. (\d)` → `U.S.C. § $1` before tokenization would address it, or the tokenizer regex could be relaxed to make `§` optional.

aseem-ai:
## Fixes
Fixes #300

## Summary
Federal Register and executive-order drafting omits the section mark, so
`50 U.S.C. 1701` extracted as nothing while `50 U.S.C. § 1701` worked.

This makes the marker optional in the existing law citation pattern rather
than adding a normalizing pre-pass. The pattern comes from reporters-db via
the `$section_marker` variable, and that variable is used by exactly one of
the 373 law regexes there — the U.S.C. one — so widening it at eyecite's
existing variable-override chokepoint touches only U.S.C. and needs no
hardcoded reporter name. A title and a section number are both still
required, so bare prose mentions of the code do not match. No lookahead is
introduced, so `HyperscanTokenizer` still compiles it.

Lettered sections like `§ 2339B` are unchanged and still unmatched; that is
#146 / #331.

## Tests
`python -m unittest discover -s tests -p 'test_*.py'` → 56 tests, OK.

New test `test_find_usc_citations_without_section_mark` fails on main
(`[] != [FullLawCitation]`) and passes here. It covers the no-§ form, the
§ form still matching once and not twice, `et seq.`, the sentence from the
issue, and four prose negatives (`the U.S.C. is a code`, `U.S.C. Title 18`,
`see U.S.C.`, `codified at 50 U.S.C.`).

Also checked by hand: extraction over `tests/assets/opinion.txt` returns the
same 170 citations before and after.

## AI Disclosure
- [x] No AI tools were used to create the content of this PR.
- [ ] Parts of this PR were created with the help of an AI tool, and I have carefully reviewed all of its content and take full responsibility for it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,10 @@ Fixes:
 - Fix 2 `TypeError`s raised in `get_citations(markup_text=...)` when
   `clean_steps` was omitted or lacked `"html"`. The documented fallback that
   prepends the `html` step was both unreachable and broken.
+- Extract U.S. Code citations written without the section mark, like
+  `50 U.S.C. 1701`. Executive-branch and Federal Register drafting omits the
+  `§`, so those cites were missed entirely. A title and a section number
+  are still both required. #300
 
 ## Current
 
--- a/eyecite/tokenizers.py
+++ b/eyecite/tokenizers.py
@@ -137,6 +137,12 @@ def _populate_reporter_extractors():
     raw_regex_variables = deepcopy(RAW_REGEX_VARIABLES)
     raw_regex_variables["full_cite"][""] = "$volume $reporter,? $page"
     raw_regex_variables["page"][""] = rf"(?P<page>{PAGE_NUMBER_REGEX})"
+    # Executive-branch drafting omits the section mark ("50 U.S.C. 1701"),
+    # so the marker is optional. Only the U.S.C. regex uses this variable,
+    # and it still requires both a title and a section number. See #300.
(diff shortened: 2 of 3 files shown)


anseljh:
@claude review

